Issue #2, opened 2018-02-05
feature request: delayed_job instrumentation


ibotty:
It would be great to also collect delayed_job metrics.

SamSaffron:
Sure, PR totally welcome here for an out-of-the-box solution like sidekiq has

mbobin:
I can have a crack at it using lifecycle callbacks around(:invoke_job). Is that OK?

SamSaffron:
Yes that seems like the way to go!

SamSaffron:
cool, this is complete now!

PR #5, opened 2018-02-12
add instrumentation for delayed job

mbobin:
Hello!

This is my attempt for the delayed_job metrics #2 issue. I'm not really convinced that `instrumentation/delayed_job` is the right place for the plugin definition, but it enables this kind of usage:

```ruby
PrometheusExporter::Instrumentation::DelayedJob.register_plugin
```

Let me know what you think.


lib/prometheus_exporter/instrumentation/delayed_job.rb
```diff
@@ -0,0 +1,42 @@
...
+      start = ::Process.clock_gettime(::Process::CLOCK_MONOTONIC)
+      result = block.call(job, *args)
+      success = true
+      result
+    ensure
+      duration = ::Process.clock_gettime(::Process::CLOCK_MONOTONIC) - start
+
+      @client.send_json(
+        type: "delayed_job",
+        name: YAML.load(job.handler).job_data["job_class"].to_s,
```
SamSaffron: oh do we need this whole yaml load thing just to get the job class?

SamSaffron:
This looks fine to me. 

Only a few minor changes

- Can you also update the README with instructions? 
- Can you include a test or 2? (I am not sure if I set the example properly with sidekiq, but we should have tests here) 

I think `PrometheusExporter::Instrumentation::DelayedJob.register_plugin` is fine, not sure where else we can put it

lib/prometheus_exporter/instrumentation/delayed_job.rb
mbobin: You're right. A regexp would do a better job.

SamSaffron:
looking good, we just need the readme updated!

SamSaffron:
Thanks heaps, failure is travis being super annoying and refusing to work with ipv6

PR #8, opened 2018-02-16
Extract Runner class from CLI script

304:
Hello! 

First, thank you for this solution! It works much better with multi-processed apps than a default prometheus-client. 

But I faced a problem, when I was integrating this gem into my codebase. I found out that it was not so easy to reuse a CLI script provided with the gem. In my case, I would like to run `prometheus-exporter` server as a daemon, but the only way I see is to copy and paste 80% the code and wrap it the way I need.  

This PR adds a new class Runner, that can be reused without extra dependencies for custom starting script implementations.

Please, let me know if I missing something here or if some changes are needed.
Thanks!


SamSaffron:
Very tidy, and very well tested,! thanks heaps. The whole verbose thing makes me think we really need a better logger strategy here, cause its a bit messy now just using stderr / out vs a proper dedicated logger.

PR #10, opened 2018-03-05
Instrument http_queue_duration_seconds by looking at X-Request-Start headers

siebertm:
This MR adds the possibility to instrument http queuing time when the upstream http server is configured to provide the `X-Request-Start` header.

I tried testing as far as I could, but had no idea on how to properly test the middleware part, as there weren't any existing tests yet... Any hints?

### DIFF
--- a/lib/prometheus_exporter/middleware.rb
+++ b/lib/prometheus_exporter/middleware.rb
@@ -23,9 +23,12 @@ class PrometheusExporter::Middleware
   end
 
   def call(env)
+    queue_time = measure_queue_time(env)
+
     MethodProfiler.start
     result = @app.call(env)
     info = MethodProfiler.stop
+
     result
   ensure
     status = (result && result[0]) || -1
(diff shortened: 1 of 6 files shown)


siebertm:
Meh, I must have some calculation errors in the code. Don't merge yet, please

lib/prometheus_exporter/middleware.rb
```diff
@@ -39,9 +42,47 @@ def call(env)
...
+    Process.clock_gettime(Process::CLOCK_REALTIME)
+  end
+
+  QUEUE_TIME_HEADERS = %w(HTTP_X_REQUEST_START HTTP_X_QUEUE_START)
+
+  # get the content of the x-queue-start or x-request-start header
+  def queue_start(env)
+    header = QUEUE_TIME_HEADERS.detect { |h| env.key?(h) }
+    value = env[header]
+    unless value.nil? || value == ''
```
SamSaffron: oh I think you can return a '' back which is not handled right by measure_queue_time

lib/prometheus_exporter/middleware.rb
```diff
@@ -39,9 +42,47 @@ def call(env)
...
+  # need to use CLOCK_REALTIME, as nginx/apache write this also out as the unix timestamp
+  def request_start
+    Process.clock_gettime(Process::CLOCK_REALTIME)
+  end
+
+  QUEUE_TIME_HEADERS = %w(HTTP_X_REQUEST_START HTTP_X_QUEUE_START)
+
+  # get the content of the x-queue-start or x-request-start header
+  def queue_start(env)
+    header = QUEUE_TIME_HEADERS.detect { |h| env.key?(h) }
```
SamSaffron: I kind of prefer 

```
value = env['HTTP_X_REQUEST_START'] || env['HTTP_X_QUEUE_START']
```

The detect seems a bit superflous here, also we have frozen strings so this should be snappy.

SamSaffron:
overall very nice change! we will use it in Discourse!

lib/prometheus_exporter/middleware.rb
siebertm: yes, that's better!

lib/prometheus_exporter/middleware.rb
siebertm: not sure what you mean here... care to elaborate?

siebertm:
With the latest changes, the values I see now in prometheus match up with what i see in scout apm. We're using it on our staging environment right now, looks good.

SamSaffron:
Thanks heaps, big fan of this change! 

Now to teach travis how to deal with ipv6 so it stops complaining 💥

PR #12, opened 2018-03-20
Fix metric names and HELP text

peterbourgon:
By convention, Prometheus counter metrics should have a suffix of `_total`, and all help strings should be proper English sentences.


SamSaffron:
completely agree with the change ❤️  it may break monitoring and so on, can you add a changelog file as well so people are aware of it?

PR #20, opened 2018-06-01
Update README to note that default_labels & default_prefix works only on

yosiat:
single process mode

I tested this only on Single Process mode which there it works fine.

About the multi-process mode: I am not sure how to implement this, I don't see a global way of adding all metrics  & instrumentation (like process/middleware), maybe sending to prometheus_exporter in addition to metric values to send "default_labels" ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,14 +324,16 @@ thing2 12
 
 ### Metrics default prefix / labels
 
+_This only works in single process mode_
+
 You can specify default prefix or labels for metrics, for example:
 
 ```ruby
 # Specify prefix for metric names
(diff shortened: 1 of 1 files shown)


SamSaffron:
cool! thanks!

yosiat:
@SamSaffron do you have any idea on how can I implement `default_labels` with multi-process support?

SamSaffron:
cool, sure this is how I would do it

https://github.com/discourse/prometheus_exporter/blob/master/bin/prometheus_exporter#L21-L23 

we need another command line option

yosiat:
The command line option won't work on my deployment.
I want to set-up one prometheus_exporter for all instances

PR #23, opened 2018-07-04
Allow empty prefix for metrics running multi process mode

terlar:
As mentioned in the documentation:
https://ruby-doc.org/stdlib-2.5.1/libdoc/optparse/rdoc/OptionParser.html#class-OptionParser-label-Type+Coercion

`String – Any non-empty string`

If you would like to skip the prefix by just providing `--prefix=''` it is not
possible as long as the Type Coercion String is specified.

Is there any other benefits to having the String coercion or would it be okay to omit it as proposed in this PR.

### DIFF
--- a/bin/prometheus_exporter
+++ b/bin/prometheus_exporter
@@ -24,7 +24,7 @@ def run
            "Timeout in seconds for metrics endpoint (default: #{PrometheusExporter::DEFAULT_TIMEOUT})") do |o|
       options[:timeout] = o.to_i
     end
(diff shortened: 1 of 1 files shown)


SamSaffron:
This looks good, failure looks unrelated, I wonder what is up with it.

terlar:
Yeah, that failure seems to randomly occur, I also experienced it when running the tests locally. Also saw the same error when you merged the other PR earlier. Will let you know if I find something, but don't feel very optimistic at this point. Thanks for the quick handling 👍

PR #24, opened 2018-07-09
Add documentation for Sidekiq process info

thejchap:


SamSaffron:
Thanks this is right!

thejchap:
Hm, actually I've been encountering some issues with the Sidekiq process stats.

When running Sidekiq via systemd and using the method in this PR to instrument, the reporter will continue to report the last RSS number after the Sidekiq process has been stopped. I am attaching images from Grafana and relevant syslog entries, but confirmed this by CURLing the reporter endpoint on the instance itself

`Jul 11 19:04:35 ip-10-1-1-236 systemd[1]: Stopped Sidekiq`

screenshot 2018-07-11 15 59 46

`Jul 11 19:49:27 ip-10-1-1-236 systemd[1]: Started Sidekiq.`

screenshot 2018-07-11 16 02 27

I will continue to look into this, just wanted to bring up in case you had any thoughts off the top @SamSaffron

SamSaffron:
Oh this is by design here: 

https://github.com/discourse/prometheus_exporter/blob/master/lib/prometheus_exporter/server/sidekiq_collector.rb

At the moment the collector has not logic regarding nuking the counters when nothing has been heard. 

I agree it makes sense to add an expiry in the collector... something like ... if you heard nothing from sidekiq for say 3 minutes, stop reporting the metrics.

thejchap:
@SamSaffron that makes sense for the sidekiq stats, I can work on implementing the expiry like you mention

What I was referring to though were the process stats (rss/etc).....those continue to be reported as the last known metric even after Sidekiq has been stopped (which I believe would be relevant to this: https://github.com/discourse/prometheus_exporter/blob/master/lib/prometheus_exporter/instrumentation/process.rb)

might need to dig a little deeper into that hook in sidekiq config to make sure its doing whats expected

SamSaffron:
The process collector actually does cycle,
https://github.com/discourse/prometheus_exporter/blob/master/lib/prometheus_exporter/server/process_collector.rb#L6
I wonder why you are having issues.

thejchap:
@SamSaffron welp not sure what happened or changed but a few Ansible runs later working on other stuff and it seems to have been fixed 🤷‍♂️ Haha thanks!

PR #25, opened 2018-07-13
Add possibility to set custom_labels in multi-process mode

4xposed:
Recently I had the need to use this feature set default labels on a multi-process setup as I wanted to have:
1 stand-alone prometheus_exporter server and many different projects reporting to that one server, while still be able to identify which metrics belong to which service

With this changes it would be possible to do the following on the different projects:
```
PrometheusExporter::Client.new(custom_labels: { service: SERVICE_NAME })
```

and on the server's `/metrics` get the following: 
```
http_requests_total{controller="home",action="index",service="CoolService",status="200"} 5
http_requests_total{controller="home",action="index",service="NotCoolService",status="200"} 2
```

### DIFF
--- a/lib/prometheus_exporter/client.rb
+++ b/lib/prometheus_exporter/client.rb
@@ -35,7 +35,7 @@ class PrometheusExporter::Client
   MAX_SOCKET_AGE = 25
   MAX_QUEUE_SIZE = 10_000
 
-  def initialize(host: 'localhost', port: PrometheusExporter::DEFAULT_PORT, max_queue_size: nil, thread_sleep: 0.5, json_serializer: nil, custom_labels: {})
+  def initialize(host: 'localhost', port: PrometheusExporter::DEFAULT_PORT, max_queue_size: nil, thread_sleep: 0.5, json_serializer: nil, custom_labels: nil)
     @metrics = []
 
     @queue = Queue.new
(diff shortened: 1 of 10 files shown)


lib/prometheus_exporter/client.rb
```diff
@@ -57,6 +57,12 @@ def initialize(host: 'localhost', port: PrometheusExporter::DEFAULT_PORT, max_qu
     @thread_sleep = thread_sleep
 
     @json_serializer = json_serializer == :oj ? PrometheusExporter::OjCompat : JSON
+
+    @custom_labels = custom_labels
```
SamSaffron: A tiny bit wasteful, let's just let custom_labels be nil and only merge if it is not nil.

SamSaffron:
overall I am fine with this, thanks! just would prefer if we only get the extra work of merging when custom labels are defined.

lib/prometheus_exporter/client.rb
4xposed: sure

4xposed:
@SamSaffron I did the changes you requested, thank for the feedback.

SamSaffron:
OK, last tiny change, can you add something to the README about it, and maybe changelog?

4xposed:
Ok. Done.

Thx

SamSaffron:
looks good to me! thanks, will bump a new version later this week.

PR #27, opened 2018-07-19
Added more useful metrics to be collected for delayed jobs

GeertZondervan:
I really like this library and the build in plugin for delayed job metrics, but I felt that the actual metrics that were being collected were a bit sparse. I've added a couple more metrics that, in my opinion, collect useful and necessary information about delayed jobs. I've added a summary of the duration it takes jobs to run (with a separation of failed and successful executed jobs), a summary of the amount of attempts it takes before jobs succeed and a counter that tells you how many jobs actually reached their max attempts.

I feel like these additions can help users monitor their delayed jobs better and let's them, for example, set alerts for when the amount of attempts needed for jobs to succeed significantly grows.

### DIFF
--- a/lib/prometheus_exporter/server/delayed_job_collector.rb
+++ b/lib/prometheus_exporter/server/delayed_job_collector.rb
@@ -14,7 +14,7 @@ module PrometheusExporter::Server
       @delayed_job_duration_seconds.observe(obj["duration"], labels)
       @delayed_jobs_total.observe(1, labels)
       @delayed_failed_jobs_total.observe(1, labels) if !obj["success"]
-      @delayed_job_max_attempts_reached.observe(1) if obj["attempts"] >= (Delayed::Worker.max_attempts)
+      @delayed_job_max_attempts_reached.observe(1) if obj["attempts"] >= obj["max_attempts"]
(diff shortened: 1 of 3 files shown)


lib/prometheus_exporter/server/delayed_job_collector.rb
```diff
@@ -40,6 +46,18 @@ def ensure_delayed_job_metrics
         @delayed_failed_jobs_total =
         PrometheusExporter::Metric::Counter.new(
           "delayed_failed_jobs_total", "Total number failed delayed jobs executed.")
+
+        @delayed_job_max_attempts_reached =
+            PrometheusExporter::Metric::Counter.new(
+                "delayed_job_max_attempts_reached", "Total number of delayed jobs that reached max attempts.")
```
SamSaffron: minor but we should keep our names consistent, see: 

https://prometheus.io/docs/practices/naming/#metric-names

Can you correct this, aside from this this is good to merge!

GeertZondervan: Will do!

GeertZondervan:
Thanks for the merge!
Can you tell me if a new release is coming any time soon, perhaps?

PR #30, opened 2018-07-29
Feature/grape instrumentation

thejchap:
This PR adds labels to [Grape](https://github.com/ruby-grape/grape) API endpoints. @SamSaffron let me know if you think this is appropriate to add in here or if its outside the scope of this gem and should be in its own module somewhere

### DIFF
--- a/README.md
+++ b/README.md
@@ -333,6 +333,12 @@ thing1 122
 thing2 12
 ```
 
+### GraphQL support
+
+GraphQL execution metrics are [supported](https://github.com/rmosolgo/graphql-ruby/blob/master/guides/queries/tracing.md#prometheus) and can be collected via the GraphQL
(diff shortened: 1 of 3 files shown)


SamSaffron:
I am very mixed on this cause I worry that if we start here where do we stop? 

Maybe we can make this a bit more flexible so you could inherit off the middleware and add the stuff you want eg add a: 

```
def get_json_payload(env)
  ...
end
```

Then we call that and you can override in your middleware to add what you wish? 

WebCollector would also need to provide a `get_default_labels` method, but with both in place you could extend and add grape or whatever special instrumentation you wanted while still leaning on core.

ZogStriP:
Do we still want/need this @SamSaffron?

SamSaffron:
Yeah I feel like a custom collector or adding some proper APIs to extend our internals if it is used in a grape scenario is a better approach here.

PR #32, opened 2018-08-03
allow gauges to increment and decrement. w/ burnettk

jasquat:
To work like https://github.com/prometheus/client_ruby#gauge. This supports use cases like keeping track of process counts if you only know when processes start and stop but don't have access to something like ps.

### DIFF
--- a/lib/prometheus_exporter/client.rb
+++ b/lib/prometheus_exporter/client.rb
@@ -21,7 +21,7 @@ class PrometheusExporter::Client
         keys: keys,
         value: value
       }
-      values.merge!(prometheus_exporter_action: prometheus_exporter_action) if prometheus_exporter_action
+      values[:prometheus_exporter_action] = prometheus_exporter_action if prometheus_exporter_action
       values
     end
 
--- a/lib/prometheus_exporter/client.rb
+++ b/lib/prometheus_exporter/client.rb
@@ -13,15 +13,28 @@ class PrometheusExporter::Client
       @type = type
     end
 
-    def observe(value = 1, keys = nil)
-      @client.send_json(
+    def standard_values(value, keys)
+      {
         type: @type,
         help: @help,
         name: @name,
         keys: keys,
         value: value
-      )
+      }
     end
+
+    def observe(value = 1, keys = nil)
+      @client.send_json(standard_values(value, keys))
+    end
+
+    def increment(value = 1, keys = nil)
+      @client.send_json(standard_values(value, keys).merge(action: :increment))
+    end
+
+    def decrement(value = 1, keys = nil)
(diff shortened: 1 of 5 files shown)


lib/prometheus_exporter/client.rb
```diff
@@ -13,15 +13,28 @@ def initialize(name:, help:, type:, client:)
...
-      )
+      }
     end
+
+    def observe(value = 1, keys = nil)
+      @client.send_json(standard_values(value, keys))
+    end
+
+    def increment(value = 1, keys = nil)
+      @client.send_json(standard_values(value, keys).merge(action: :increment))
```
SamSaffron: to avoid the "merge" and throwaway object I would just have standard values accept the action optionally, if nothing specified it should not add an action

lib/prometheus_exporter/metric/gauge.rb
```diff
@@ -22,5 +22,15 @@ def metric_text
     def observe(value, labels = {})
       @data[labels] = value
     end
+
+    def increment(value, labels = {})
```
SamSaffron: isn't the default 1 here?

lib/prometheus_exporter/server/collector.rb
```diff
@@ -46,7 +46,12 @@ def process(str)
           if !metric
             metric = register_metric_unsafe(obj)
           end
-          metric.observe(obj["value"], obj["keys"])
+          action = obj["action"]
+          if action && supported_actions.include?(action) && metric.respond_to?(action)
```
SamSaffron: I think we can avoid the meta programming here and just be explicit

SamSaffron:
overall I am fine with this change... we should aim for full parity with: 

https://github.com/prometheus/client_golang/blob/master/prometheus/gauge.go#L32-L53


So I guess we are missing set to current time ... but that is easy just to do by hand.

lib/prometheus_exporter/client.rb
```diff
@@ -13,15 +13,15 @@ def initialize(name:, help:, type:, client:)
...
+    def standard_values(value, keys, prometheus_exporter_action = nil)
       values = {
         type: @type,
         help: @help,
         name: @name,
         keys: keys,
         value: value
       }
-      values.merge!(action: action) if action
+      values.merge!(prometheus_exporter_action: prometheus_exporter_action) if prometheus_exporter_action
```
SamSaffron: oh no need to call merge here:

```
values[:prometheus_exporter_action] = prometheus_exporter_action if prometheus_exporter_action
```

SamSaffron:
looks good except for that one tiny change

SamSaffron:
Thanks, change looks good to me!

PR #34, opened 2018-09-23
Add support for Puma metrics

terlar:
This implementation groups them all together when running several
workers. One might potentially want to track them separately by putting
different labels. However I haven't personally had a use for it, so
currently it is all grouped together.

I've been running this code in production for a few months and it seems to be working good. Hope it can be useful to someone else.


SamSaffron:
This looks great, but we should follow the Prometheus naming conventions per: https://prometheus.io/docs/practices/naming/

Also can you add a note to the readme here regarding usage?

terlar:
Ah, thanks for the feedback, I did this before the Prometheus naming convention update and then I forgot that it was made. Will address this as well as some usage example in the readme.

terlar:
@SamSaffron I have updated the PR accordingly, are these metric names good? Also ensured it was included in the server/instrumentation files. I also added a usage section in the readme.

SamSaffron:
This is great! cc @schneems / @nateberkopec

drewish:
Kind of a nit pick but the stats are only available in 3.12+. Any objection to a PR to update the docs and call that out?

drewish:
Also when running two puma workers I'm getting a warning:
```
[1] Puma starting in cluster mode...
[1] * Version 3.12.0 (ruby 2.4.5-p335), codename: Llamas in Pajamas
[1] * Min threads: 1, max threads: 1
[1] * Environment: development
[1] * Process workers: 2
[1] * Preloading application
[1] * Listening on tcp://0.0.0.0:3000
[1] ! WARNING: Detected 2 Thread(s) started in app boot:
[1] ! #<Thread:0x0000562e31cb9480@/usr/local/bundle/bundler/gems/prometheus_exporter-3c889b088fa6/lib/prometheus_exporter/instrumentation/process.rb:7 sleep> - /usr/local/bundle/bundler/gems/prometheus_exporter-3c889b088fa6/lib/prometheus_exporter/instrumentation/process.rb:15:in `sleep'
[1] ! #<Thread:0x0000562e31c79fd8@/usr/local/bundle/bundler/gems/prometheus_exporter-3c889b088fa6/lib/prometheus_exporter/client.rb:148 sleep> - /usr/local/bundle/bundler/gems/prometheus_exporter-3c889b088fa6/lib/prometheus_exporter/client.rb:151:in `sleep'
[1] Use Ctrl-C to stop
[1] - Worker 0 (pid: 14) booted, phase: 0
[1] - Worker 1 (pid: 25) booted, phase: 0
```
Is that a problem or by design?

I'm using the suggested snipped in my `config/puma.rb`:
```rb
after_worker_boot do
  PrometheusExporter::Instrumentation::Puma.start
end
```

lib/prometheus_exporter/server/puma_collector.rb
```diff
@@ -0,0 +1,48 @@
...
+            g.observe(v, labels)
+          end
+        end
+      end
+
+      metrics.values
+    end
+
+    def collect(obj)
+      @puma_metrics << obj
```
SamSaffron: This is an ENORMOUS GIANT issue, it is leaking memory forever and never clearing stuff like process collector does. 

It urgently urgently needs redoing to follow the same pattern in process collector.

terlar: Sorry about this, if I remember correctly this was based on the process one, which does have slightly different logic, where it cleans up metrics:
```
      now = ::Process.clock_gettime(::Process::CLOCK_MONOTONIC)

      obj["created_at"] = now

      @process_metrics.delete_if do |current|
        obj["pid"] == current["pid"] || (current["created_at"] + MAX_PROCESS_METRIC_AGE < now)
      end
      @process_metrics << obj
```

I think I missunderstood the impact of this at the time and we actually ran into this memory leak at work. There was someone who fixed it, but unfortunately never contributed it, perhaps due to confusion when our company dropped support for ruby. Once again sorry for that.

This is what our current implementation looks like:
```ruby
# frozen_string_literal: true

# collects stats from puma
module PrometheusExporter
  module Instrumentation
    class Puma
      def self.start(client: nil, frequency: 30)
        puma_collector = new
        Thread.new do
          loop do
            begin
              metric = puma_collector.collect
              client.send_json metric
            rescue StandardError => e
              STDERR.puts("Prometheus Exporter Failed To Collect Puma Stats #{e}")
            ensure
              sleep frequency
            end
          end
        end
      end

      def collect
        metric = {}
        metric[:type] = 'puma'
        collect_puma_stats(metric)
        metric
      end

      def collect_puma_stats(metric)
        stats = JSON.parse(::Puma.stats)

        if stats.key? 'workers'
          metric[:phase] = stats['phase']
          metric[:workers_total] = stats['workers']
          metric[:booted_workers_total] = stats['booted_workers']
          metric[:old_workers_total] = stats['old_workers']

          stats['worker_status'].each do |worker|
            collect_worker_status(metric, worker['last_status'])
          end
        else
          collect_worker_status(metric, stats)
        end
      end

      private

      def collect_worker_status(metric, status)
        return if status.empty?

        metric[:request_backlog_total] ||= 0
        metric[:running_threads_total] ||= 0
        metric[:thread_pool_capacity_total] ||= 0
        metric[:max_threads_total] ||= 0

        metric[:request_backlog_total] += status["backlog"]
        metric[:running_threads_total] += status["running"]
        metric[:thread_pool_capacity_total] += status["pool_capacity"]
        metric[:max_threads_total] += status["max_threads"]
      end
    end
  end
end
```

I see there is another fix more along the lines of the process one, but I could update with this kind of solution if you think it is better.

Once again sorry for the inconvenience.

eviltrout: No worries, these kinds of things happen!

Upgrading the puma collector is less urgent now, but I think it might be a nice PR? We can compare to the old one and see if it's more valuable data (I suspect it is.)

PR #38, opened 2018-10-08
Use actual job class instead of wrapper

onemanstartup:
Without that patch I get only 'ActiveJob::QueueAdapters::SidekiqAdapter::JobWrapper' metrics.

### DIFF
--- a/lib/prometheus_exporter/instrumentation/sidekiq.rb
+++ b/lib/prometheus_exporter/instrumentation/sidekiq.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module PrometheusExporter::Instrumentation
   class Sidekiq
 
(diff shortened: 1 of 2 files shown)


onemanstartup:
While I am here. How do you look on adding readers or accessors to client and remote metric?
```ruby
  class PrometheusExporter::Client
    attr_reader :metrics
    class RemoteMetric
      attr_accessor :name
    end
  end
```
I want to write this code without monkeypatching.

```ruby
  # To use as Metrics.awesome.observe(1)
  module Metrics
    class << self
      def client
        PrometheusExporter::Client.default
      end

      def method_missing(method, args=nil)
        client.metrics.find { |m| m.name == method.to_s }
      end
    end
  end
```

SamSaffron:
I think the patch looks good to me! can you add  `# frozen_string_literal: true` to the file and make sure the exporter is tested? 

Happy to add readers and stuff that makes your code less hacky.

onemanstartup:
@SamSaffron I don't know if it is the right place to add test :) But here you go, added test, string literal and squashed

SamSaffron:
Yep, looks correct to me! Thanks!

PR #39, opened 2018-10-15
add histogram metric type

jstern:
We have a polyglot set of services for which we've already established a pattern of collecting certain metrics as histograms, and I'd like to be able to use this gem to collect the equivalent metrics from the ruby apps in that ecosystem as well.

This is a pretty naive implementation, so please let me know if there are changes you'd like to see in order to merge this.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,10 +61,12 @@ server.start
 gauge = PrometheusExporter::Metric::Gauge.new("rss", "used RSS for process")
 counter = PrometheusExporter::Metric::Counter.new("web_requests", "number of web requests")
 summary = PrometheusExporter::Metric::Summary.new("page_load_time", "time it took to load page")
+histogram = PrometheusExporter::Metric::Histogram.new("api_access_time", "time it took to call api")
 
 server.collector.register_metric(gauge)
 server.collector.register_metric(counter)
 server.collector.register_metric(summary)
+server.collector.register_metric(histogram)
 
 gauge.observe(get_rss)
 gauge.observe(get_rss)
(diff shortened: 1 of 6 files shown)


SamSaffron:
Overall this looks great to me! Do we want to allow people to pick buckets when they initialize the metric? It feels to me like something people will want. 

Can you also update the README with some info about this? 

Can you change this as well? I think we always target Rubies with `+` support

```
(String.new << "a").encoding
=> #<Encoding:ASCII-8BIT>

((+"") << "a").encoding
=> #<Encoding:UTF-8>
```

jstern:
Thanks Sam! 

I think letting the user pick the buckets should be pretty straightforward, but I was wondering if you had thoughts on what that should look like when using `client.register(:histogram, ...)`?

I'll make the string change but the same issue already exists in the summary class I mostly copied this from. Maybe those other places where this exists should be touched in a separate PR?

I'll add examples to the readme when the rest is ironed out.

SamSaffron:
I pushed a fix for summary, thanks for that! Interestingly we don't do custom quantiles for summary, we should add support for that. 

I think if you are using client.register ... you should just rely on defaults, anything fancy really should use custom transport imo, it is far more efficient to move data around like that.

Happy to merge this in as soon as you are ready, maybe just add the readme. 

Adding support for custom quantiles in summary now so we have better parity at least in the api level

jstern:
@SamSaffron - re client.register, happy to use defaults. So far we haven't had a reason to specify our own buckets. 

Note that I I did change the constructor for histogram from 

```ruby
# original
Histogram('foo', 'foo time', [1, 2, 3])
```

to this using an `opts` dict

```ruby
# latest
Histogram('foo', 'foo time', buckets: [1, 2, 3])
```

I think if you ever changed your mind about allowing options in client.register this will let you have a more uniform api for passing them. But let me know if you prefer the "original" version above and I can change it back along with the readme examples.

Otherwise, I'm happy with this if you are.

SamSaffron:
This looks good to me! We should do the same for summary then so we are consistent. Can you do a PR to update the readme? 

Very happy with the change, thank you 👍

PR #40, opened 2018-10-16
Register PumaCollector in Server

rickenharp:
While trying to use the Puma collector with the `prometheus_exporter` binary, it kept throwing exceptions every time the Puma collector sent stats to /send-metrics.
I was able to fix it by adding the PumaCollector to the registered collectors.


SamSaffron:
Thanks heaps!

PR #41, opened 2018-10-18
mention custom quantiles and buckets in readme

jstern:
@SamSaffron - here's a readme update describing the custom quantiles and buckets. Please let me know if you'd like me to change summary to use an opts hash, or change histogram to use a buckets arg. I feel like they should be consistent but happy to go with whichever consistent approach feels right to you.


SamSaffron:
Thanks! looks good

PR #42, opened 2018-10-22
use the opts pattern for custom quantiles too

jstern:
Hi @SamSaffron 

I think this got lost in the shuffle with my last PR. Let me know if you prefer to go in the other direction and use a positional args for histogram; happy to make that change instead if that's your preference.

Can you let me know when you think you might be ready to publish a new version with histogram support on rubygems.org, and if there's anything else I can help with to speed that along?

Thanks!


SamSaffron:
Change looks perfect to me, will push a version today

SamSaffron:
OK.. 0.4.0 is now out!

jstern:
Awesome, thank you :)

PR #43, opened 2018-11-03
Add support for Hutch

drewish:
I based this on the Sidekiq instrumentation and used [Hutch's message processing tracer interface](https://github.com/gocardless/hutch#message-processing-tracers).


SamSaffron:
I am happy to merge this but think we need to take a step back and think about autoloading this stuff as needed vs bringing in all instrumentation in one go

drewish:
Yeah I think having people opt in to each type of monitoring they want would make sense. I guess that could just be requiring each instrumentation manually? You have thoughts on how less auto loading on server side could operate?

SamSaffron:
not 100% sure if we should go with explicitly requiring people type requires for every instrumentation vs a simple autoload trick with constants (which would automatically work in the collector if we design it right) 

For now it is something to think about

PR #44, opened 2018-11-03
MySQL2 support

drewish:
I'm not 100% sure I got all the methods but these are the obvious ones that jumped out when reviewing [the readme](https://github.com/brianmario/mysql2).

### DIFF
--- a/lib/prometheus_exporter/middleware.rb
+++ b/lib/prometheus_exporter/middleware.rb
@@ -19,6 +19,11 @@ class PrometheusExporter::Middleware
           :exec, :async_exec, :exec_prepared, :send_query_prepared, :query
         ], :sql)
       end
+      if defined? Mysql2::Client
+        MethodProfiler.patch(Mysql2::Client, [:query], :sql)
+        MethodProfiler.patch(Mysql2::Statement, [:execute], :sql)
+        MethodProfiler.patch(Mysql2::Result, [:each], :sql)
+      end
     end
   end
 


SamSaffron:
I think you are intercepting too much stuff here, see: 

https://github.com/MiniProfiler/rack-mini-profiler/blob/master/lib/patches/db/mysql2.rb

looks like each and query is all we need.

drewish:
That makes sense. I’ll hook that up and give it a whirl. I think you do still want the execute on the statement though.

drewish:
That seems to do a reasonable job. Here's some metrics I'm seeing:
```
# HELP ruby_http_sql_duration_seconds Time spent in HTTP reqs in SQL in seconds.
# TYPE ruby_http_sql_duration_seconds summary
ruby_http_sql_duration_seconds{controller="settings",action="index",quantile="0.99"} 0.004706399995484389
ruby_http_sql_duration_seconds{controller="settings",action="index",quantile="0.9"} 0.004706399995484389
ruby_http_sql_duration_seconds{controller="settings",action="index",quantile="0.5"} 0.004706399995484389
ruby_http_sql_duration_seconds{controller="settings",action="index",quantile="0.1"} 0.004706399995484389
ruby_http_sql_duration_seconds{controller="settings",action="index",quantile="0.01"} 0.004706399995484389
ruby_http_sql_duration_seconds_sum{controller="settings",action="index"} 0.004706399995484389
ruby_http_sql_duration_seconds_count{controller="settings",action="index"} 1
ruby_http_sql_duration_seconds{controller="templates",action="custom_templates",quantile="0.99"} 0.008353300007001963
ruby_http_sql_duration_seconds{controller="templates",action="custom_templates",quantile="0.9"} 0.008353300007001963
ruby_http_sql_duration_seconds{controller="templates",action="custom_templates",quantile="0.5"} 0.008353300007001963
ruby_http_sql_duration_seconds{controller="templates",action="custom_templates",quantile="0.1"} 0.008353300007001963
ruby_http_sql_duration_seconds{controller="templates",action="custom_templates",quantile="0.01"} 0.008353300007001963
ruby_http_sql_duration_seconds_sum{controller="templates",action="custom_templates"} 0.008353300007001963
ruby_http_sql_duration_seconds_count{controller="templates",action="custom_templates"} 1
```
And here's the timing that ActiveRecord is showing:
```
Started GET "/sites/2070563435179181053/settings.json?country=US&currency=USD" for 172.18.0.16 at 2018-11-05 13:06:41 +0000
Processing by SettingsController#index as JSON
Completed 200 OK in 92ms (Views: 2.0ms | ActiveRecord: 4.5ms)

Started GET "/sites/2070563435179181053/templates/custom_templates.json?locale=en-US" for 172.18.0.16 at 2018-11-05 13:06:41 +0000
Processing by TemplatesController#custom_templates as JSON
Completed 200 OK in 38ms (Views: 15.4ms | ActiveRecord: 8.2ms)
```

drewish:
@SamSaffron any other thoughts on this?

SamSaffron:
do we need statment execute? is it missing from rack mini profiler? I am happy to merge this but if this is broken in mini profiler we should fix that too.

drewish:
It looked like a distinct path to execute a query that didn’t call into the connection to me. https://github.com/brianmario/mysql2/blob/master/lib/mysql2/statement.rb calls https://github.com/brianmario/mysql2/blob/master/ext/mysql2/statement.c which seems to reference the client only in the extension code and not in Ruby land.

drewish:
I meant to add: I’m happy to put together a little test case to try it out. I’m not sure if ActiveRecord uses the prepared statements so it might be something that isn’t widely used and therefore easy to miss.

SamSaffron:
I see, I am fine with this, I guess we need to add this to mini profiler.

PR #46, opened 2018-11-30
allow passing opts to Histogram and Summary from client. w/ burnettk

jasquat:
This will allow clients to change buckets and quantiles for histograms and summaries.

### DIFF
--- a/lib/prometheus_exporter/client.rb
+++ b/lib/prometheus_exporter/client.rb
@@ -4,16 +4,16 @@ require 'socket'
 require 'thread'
 
 module PrometheusExporter
-
   class Client
     class RemoteMetric
       attr_reader :name, :type, :help
 
-      def initialize(name:, help:, type:, client:)
+      def initialize(name:, help:, type:, client:, opts: {})
         @name = name
         @help = help
         @client = client
         @type = type
+        @opts = opts
       end
 
       def standard_values(value, keys, prometheus_exporter_action = nil)
@@ -39,7 +39,6 @@ module PrometheusExporter
       def decrement(keys = nil, value = 1)
         @client.send_json(standard_values(value, keys, :decrement))
       end
-
     end
 
     def self.default
@@ -83,8 +82,8 @@ module PrometheusExporter
       @custom_labels = custom_labels
     end
 
-    def register(type, name, help)
-      metric = RemoteMetric.new(type: type, name: name, help: help, client: self)
+    def register(type, name, help, opts = {})
+      metric = RemoteMetric.new(type: type, name: name, help: help, client: self, opts: opts)
       @metrics << metric
       metric
     end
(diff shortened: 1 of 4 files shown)


hoffoo:
can a maintainer look at this, i'd love to see it merged

lib/prometheus_exporter/client.rb
```diff
@@ -6,18 +6,20 @@
 class PrometheusExporter::Client
 
   class RemoteMetric
-    def initialize(name:, help:, type:, client:)
+    def initialize(name:, help:, type:, client:, opts:)
```
SamSaffron: why is opts not optional here?

lib/prometheus_exporter/server/collector.rb
```diff
@@ -78,6 +78,7 @@ def register_metric(metric)
     def register_metric_unsafe(obj)
       name = obj["name"]
       help = obj["help"]
+      opts = symbolize_keys(obj["opts"] || {})
```
SamSaffron: Do we need symbolize keys here? If so why, also why run a bunch of code if not opts were given?

SamSaffron:
@hoffoo / @jasquat  sure I left some comments, this would need a rebase and once comments are addressed we can look at merging it in.

lib/prometheus_exporter/server/collector.rb
jasquat: I think we were symbolizing keys because that's what the summary and histogram classes are looking for: https://github.com/discourse/prometheus_exporter/blob/master/lib/prometheus_exporter/metric/histogram.rb#L10. If people are using those models we would break them by changing the options hash to use string keys instead of symbols. So it seems better to continue symbolizing the keys.

This could be moved down to only happen in the summary and histogram "when" blocks but it seems like de-duping code is more advantagious here since running inject on a blank hash shouldn't really do anything.

lib/prometheus_exporter/client.rb
jasquat: No reason. Will update to be optional.

lib/prometheus_exporter/client.rb
```diff
@@ -4,16 +4,16 @@
 require 'thread'
 
 module PrometheusExporter
-
   class Client
     class RemoteMetric
       attr_reader :name, :type, :help
 
-      def initialize(name:, help:, type:, client:)
+      def initialize(name:, help:, type:, client:, opts: {})
```
SamSaffron: `opts: nil`

Then `opts || {}` is not needed so we allocate one less hash.

burnettk: if the caller explicitly passes in nil to initialize, this would still be necessary to ensure opts is a hash.

SamSaffron: not until way later, this check can be done in the collector, the remote metric can simply not send that field across if it is nil. (at least I think) :thinking:

jasquat: yeah, defaulting to nil now and setting only in the collector. we also added some tests for it. i think the test failure in travis was intermittent. this pull request is ready from our perspective. we can squash the commits if you want.

SamSaffron:
This is looking good! can you do a followup PR to update the readme?

PR #47, opened 2018-12-02
Add active record support

blackjid:
Hi!, I've been working to add metrics from sql on projects that use active record. We use it on rails.

- It uses the rails instrumentation api to hook to sql queries notifications
- It normalizes the queries so we can group them together and have labels with the query form added to the metric
- It skips cached and schema queries

image

I hope this is useful for more people. It's been for us.


lib/prometheus_exporter/server/active_record_collector.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+
+    def collect(obj)
+      default_labels = { query: obj["query"], action: obj["action"] }
+      custom_labels = obj["custom_labels"]
+      labels = custom_labels.nil? ? default_labels : default_labels.merge(custom_labels)
+
+      ensure_active_record_metrics
+      @active_record_query_duration_seconds.observe(obj["duration"], labels)
+      @active_record_query_duration_seconds_summary.observe(obj["duration"], labels)
```
blackjid: I'm thinking of removing this query from the summary metric. Probably keep the action label.

@lcasassa do you think it makes sense the add the action label to the summary metric?? The action label tells us the kind of query, for example `User Load`, `User Update`.

SamSaffron:
We need to extend the documentation here a ton and make sure there are some switches for fidelity here. 

My big concern here is that this has very strong potential to blow prometheus storage especially in cases where you can not normailze eg: `User.joins('join user_extra on id = #{user.id}')`

lib/prometheus_exporter/utils/sql_sanitizer.rb
```diff
@@ -0,0 +1,109 @@
...
+
+      def has_encodings?(encodings=['UTF-8', 'binary'])
+        encodings.all?{|enc| Encoding.find(enc) rescue false}
+      end
+
+      def scrubbed(str)
+        return '' if !str.is_a?(String) || str.length > MAX_SQL_LENGTH # safeguard - don't sanitize or scrub large SQL statements
+        return str if str.valid_encoding? # Whatever encoding it is, it is valid and we can operate on it
+
+        if str.respond_to?(:scrub) # Prefer to scrub before we have to convert
```
SamSaffron: scrub should be in place ... it has been around since 2.0 we require 2.3

lib/prometheus_exporter/utils/sql_sanitizer.rb
```diff
@@ -0,0 +1,109 @@
+require 'prometheus_exporter/utils/rails'
+require 'active_support/core_ext/string/filters'
+
+module PrometheusExporter
+  module Utils
+    class SqlSanitizer
```
SamSaffron: I think we need a bunch of tests for the sanitizer it also would help clarify how it works

lib/prometheus_exporter/utils/sql_sanitizer.rb
blackjid: that makes sense

lib/prometheus_exporter/utils/sql_sanitizer.rb
blackjid: I didn't want to reinvent the wheel so the code from the sanitizers was taken from the Scout App apm gem. They have tests for the sanitizers.

But, looking into their repo, I just realize reading the License file that their code is not open source :( So I don't know how to proceed.

blackjid:
Hey @SamSaffron, my experience using Prometheus is very limited so the feedback is very welcomed.

About opting-in, I was planning to conditionally initialize this instrument base on an environmental variable on my project.

cschneid:
Thanks for the ping - I can go look into this. Our agent isn't open source, but we like to share code when possible. I'm asking internally to figure out what the policy and approach would be for this.

blackjid:
closing this... :)

PR #48, opened 2018-12-04
Check for custom labels in the generic metrics

oestrich:
Hello, we're trying to get custom labels on the client for the generic metrics (gauges, counters, etc) and the custom labels option being sent over appeared to be ignored for those.

This fixes that by merging them with the `keys` being sent over, giving preference for anything inside of `keys` over `custom_labels`.


SamSaffron:
looks good to me, seems consistent with how it is used elsewhere.

PR #51, opened 2018-12-20
Fix a typo in prometheus exporter

bpaquet:


SamSaffron:
oh my ... great catch ... we should add a test

PR #54, opened 2019-02-12
Identify processes by PID and type

### LINKED ISSUE #53: Process metrics get lost on docker/kubernetes environments
We use prometheus_exporter in a kubernetes deployment where multiple containers (puma, delayed_job, ..) report to the same exporter. In this environment, it is common that the main processes in the different containers all get PID 1.

Because of https://github.com/discourse/prometheus_exporter/blob/fb41e20d6fe800ca86ce16b395ad6c1256228b3b/lib/prometheus_exporter/server/process_collector.rb#L65, only one set of process metrics may exist per PID. Hence, depending on which container reported last, we only see those process metrics. The ones of the other containers are lost.

I think a check on `(obj["pid"] == current["pid"] && obj["process_type"] == current["process_type"])` would resolve this issue, as we assign different `process_types` per container.

codez:
Fixes #53

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,8 +179,10 @@ end
 # in unicorn/puma/passenger be sure to run a new process instrumenter after fork
 after_fork do
   require 'prometheus_exporter/instrumentation'
-  PrometheusExporter::Instrumentation::Process.start(type:"web")
+  PrometheusExporter::Instrumentation::Process.start(type: "web")
 end
+
+When monitoring process metrics, each process must have an unique type/PID pair indentifying the process. This might be an issue in docker environments where primary processes all get PID 1.
 
 ```
 
--- a/lib/prometheus_exporter/server/process_collector.rb
+++ b/lib/prometheus_exporter/server/process_collector.rb
@@ -62,7 +62,8 @@ module PrometheusExporter::Server
       obj["created_at"] = now
 
       @process_metrics.delete_if do |current|
-        obj["pid"] == current["pid"] || (current["created_at"] + MAX_PROCESS_METRIC_AGE < now)
+        (obj["pid"] == current["pid"] && obj["process_type"] == current["process_type"]) ||
+        (current["created_at"] + MAX_PROCESS_METRIC_AGE < now)
       end
       @process_metrics << obj
     end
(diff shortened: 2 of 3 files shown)


SamSaffron:
Thinking about this a bit more maybe this should be optional opt in. 

Technically old pid 77 could be web, and new pid 77 could be sidekiq all in the same container. This will amend the behavior of this edge case. 

Maybe for a multi container "collector" opt in for multi container in the initializer. Type is not enough here imo, you would want to pass in some container id for it to be correct.

codez:
So what do you think of introducing an optional 'container_id' that could be passed to the process instrumentation and that would be used as an identifier, together with the pid, by the process collector if present?

Only using pid and container_id as identifier would not work for the edge case you mentioned. Should type be included as well?

silvermind:
I have a similar issue, with the Process Collector in a multi container multi host setup, where I want one Collector Container per Host (identified by labels).

In a Rails Initializer i have a Client with defined custom_labels:
```
PrometheusExporter::Client.default = PrometheusExporter::Client.new(custom_labels: { container_name: container_name })
```

the logic ```obj["pid"] == current["pid"]```also remove those custom labels.
All other collectors have the following code to ensure the labels survive the collector:

```
default_labels = {}
custom_labels = obj['custom_labels']
labels = custom_labels.nil? ? default_labels : default_labels.merge(custom_labels)
```

Having custom labels assign at the client source, may so solve your problem @codez 

something like:
```metric_key = { pid: m["pid"], type: m["process_type"]}.merge(m['custom_labels']) ```
instead of: https://github.com/discourse/prometheus_exporter/blob/fd8404fc3b82f450ea81bdfe483a17c113716fa1/lib/prometheus_exporter/server/process_collector.rb#L37

What do you think about *custom_labels* in the Process Collector @SamSaffron ?

SamSaffron:
how does this look? 

we now automatically scope on hostname and allow you to add custom labels if you wish.

kevine-voicera:
@SamSaffron @codez @silvermind Is there a way to effectively pass down the original pod that was supplying the metric to the collector? Right now they all identify as the collector's pod.

PR #56, opened 2019-02-20
allow client queue to be flushed

PaulBabka:
Thanks for this product.  Just what we need.

We have a number of really short-lived processes (cron / scripts) than finish before the worker thread can send the queue.  Took me a while to figure out why.

I've added a 'flush' methods that encourages ruby to allow the worker thread process time to send any outstanding queued entries.  This seems to do the trick.  I'm an architect (not a developer), so any feedback on my code and test would be most humbly welcomed.

FYI

We have 2 use cases for prometheus_exporter:
1) Multi-threaded instrumentation of Rails application (passenger) where even if Prometheus ruby client multi-threaded issue is resolved, we don't want to expose metrics in-band as part of the application.  Sending metrics to prometheus_client, and exporting on an alternative port is perfect.
2) Short-running cron/scripts where push-gateway or Node exporter textfile is inappropriate/inadvisable.

Cheers

### DIFF
--- a/test/client_test.rb
+++ b/test/client_test.rb
@@ -1,0 +1,30 @@
+require "test_helper"
+require 'prometheus_exporter/client'
+
+# Mock queue reduces its length as it is checked
+class MockQueue
+  def length
+    @queue_length = (@queue_length || 3) - 1
+    @queue_length < 0 ? 0 : @queue_length
+  end
+end
+
+describe PrometheusExporter::Client do
+  
+  before do
+    @client = PrometheusExporter::Client.default
+  end
+
+  describe "flush" do
+    it "sends all queued entries within the timeout" do
+      @client.instance_variable_set(:@queue, MockQueue.new())
+      assert @client.flush(0.1)
+    end
+    it "fails to send queued entries before the timeot" do
+      @client.instance_variable_set(:@queue, [99])
+      refute @client.flush(0.1)
+    end
+  end
+  
+end
+  
--- a/test/client_test.rb
+++ b/test/client_test.rb
@@ -18,11 +18,11 @@ describe PrometheusExporter::Client do
   describe "flush" do
     it "sends all queued entries within the timeout" do
       @client.instance_variable_set(:@queue, MockQueue.new())
-      assert @client.flush(0.1)
+      assert @client.flush(0.01)
     end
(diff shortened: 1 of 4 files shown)


SamSaffron:
This is fine in principal but we don't need the Timeout here, you just track when you started the call to flush and run a simple loop. 

I really dislike Timeout cause it is a pretty unfriendly pattern with Thread.kill and all and we don't need it here.

PaulBabka:
Hi Sam.  Timeout removed, but still keeping an escape mechanism in the event the queue cannot be emptied in a timely fashion.  Otherwise this method would block forever in the event that prometheus_exporter server was offline and messages could not be sent.

Also - would you be amenable to a pull request to include an info metric containing the version?  I will have lots of prometheus_exporters running around on various servers, and knowing the version is always handy when troubleshooting.  Something like:

# HELP ruby_collector_info prometheus_exporter info
# TYPE ruby_collector_info gauge
ruby_collector_info{version="x.x.x"} 1

You time appreciated.

Babs

test/client_test.rb
```diff
@@ -0,0 +1,30 @@
...
+describe PrometheusExporter::Client do
+  
+  before do
+    @client = PrometheusExporter::Client.default
+  end
+
+  describe "flush" do
+    it "sends all queued entries within the timeout" do
+      @client.instance_variable_set(:@queue, MockQueue.new())
+      assert @client.flush(0.01)
```
SamSaffron: This is potentially a heisentest, cause sleep 0.001 is not a guaranteed to take 1ms.

Also ... ` sleep 0.001` is an implementation detail, no need to test that. 

Instead.... I would do an "honest" flush in a test, pause client, queue a bunch of messages, unpause/flush, check it is empty.

SamSaffron:
> would you be amenable to a pull request to include an info metric containing the version?

Fine with making it an optional thing default off.

test/client_test.rb
PaulBabka: Can do.
Do you prefer a series of discreet PR (1 for flush, 1 for the info).  That way you can pick/choose?
Thanks

SamSaffron:
discreet PRs is totally the way to go.

SamSaffron:
This sadly got stale, @PaulBabka feel free to reopen discrete PRs here.

PR #57, opened 2019-02-21
Differentiate between restarted, failed and dead jobs

alexmorten:
Especially for alerting, it would be quite useful for us, to have a differentiation between jobs that were pushed back to the Redis-queue because Sidekiq was shut down (i.e. during a deployment), jobs that failed but will be retried, and jobs that exhausted all their retries.

When I was working on the `sidekiq_restarted_jobs_total ` metric, I noticed that, depending on timing, the client might or might not send messages to the collector before the process quits. For that reason, I added a wait with timeout when stopping the client.

I'm not very happy with the way I added the death_handler in `instrumentation/sidekiq.rb`. This can lead to unexpected behaviour if the client given to `PrometheusExporter::Instrumentation::Sidekiq#initialize` on each job-run changes. I would love to see suggestions for how this can be solved more elegantly.


SamSaffron:
Looking quite promising! Keep in mind weekend is about to land, this is a pretty big change so it may take a few more days to merge.

Overall I opted for less magic here ... so we init sidekiq with:

```
Sidekiq.configure_server do |config|
   config.server_middleware do |chain|
      chain.add PrometheusExporter::Instrumentation::Sidekiq
   end
end
```

So I would say you just introduce the helper into this config block with something like :

```
config.death_handlers <<  PrometheusExporter::Instrumentation::Sidekiq.death_handler 
```

This can simply use `PrometheusExporter::Client.default` 

Can you do something like this, also please update the README here explaining the changes and the changlog to note them. I noticed dead jobs don't have a duration, don't they take time to die?

alexmorten:
@SamSaffron Thank you! That's much better.

>I noticed dead jobs don't have a duration, don't they take time to die?

Can you elaborate on that? I'm not sure I understand. 
In my mind a job run has a duration, but a dead job is not necessarly bound to a single run.

SamSaffron:
I was just concerned that it takes time for a job to die and that time is not being measured. That said it is so nit picky that its probably not worth addressing, the change looks great! merging it in.

Issue #62, opened 2019-03-29
Explicitly setting counter value


dpetersen:
I have a question about a change I'd like to make but wanted to check if it would even be accepted.

I'm trying to instrument our application's Sidekiq info. I want most of what's available from `Sidekiq::Stats`, like queue lengths and processed/failed. Queues are clearly gauges, but processed/failed seems more suitable for counters since they can be reset to zero in the Sidekiq UI but in reality are always increasing.

I can only get the current processed/failure count from Sidekiq, but the API for the `PrometheusExporter::Metric::Counter` only allows the count to be incremented. Unless I resort to `instance_variable_get`/`set` shenanigans or subclass the `PrometheusExporter::Metric::Base` myself, I don't think I can accomplish what I want. I would like to be able to set the value of a counter directly.

Neither the Ruby client nor the Golang client allows this, either, so perhaps I'm an idiot and instrumenting thing incorrectly.

Would you accept a PR to allow direct setting of a Counter's value, and if so what would be a reasonable method name? I was thinking:
```ruby
def reset(to = 0, labels = {})
  # blah
end
```

Thanks for this gem. It took a while to wrap my head around how to run it in our Kubernetes environment, but the architecture works great and improves upon our old solution.

SamSaffron:
Yeah ability to reset a Counter via api sounds fine to me.

PR #63, opened 2019-04-01
Add reset method to Counter

dpetersen:
This is the feature we discussed in #62, giving me the ability to explicitly reset a Counter's value.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,12 +1,17 @@
-0.4.5 - 22-02-2019
+0.4.6 - 02-04-2019
 
+- Feature: Allow resetting a counter
 - Feature: Add sidekiq metrics: restarted, dead jobs counters
 - Fix: Client shutting down before sending metrics to collector
 
-0.4.4 - 14-02-2019
+0.4.5 - 14-02-2019
 
 - Feature: Allow process collector to ship custom labels for all process metrics
 - Fix: Always scope process metrics on hostname in collector
+
+0.4.4 - 13-02-2019
+
+- Feature: add support for local metric collection without using HTTP 
 
 0.4.3 - 11-02-2019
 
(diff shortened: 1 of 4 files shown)


SamSaffron:
Sounds good to me, can you do a quick pass on the readme+changelog and update as well, then I will merge!

dpetersen:
While looking at the `CHANGELOG`, I think what's listed there is not accurate.

While setting up this gem in my Rails app, I noticed that the `PrometheusExporter::Instrumentation::Sidekiq.death_handler` method listed in the `README` on GitHub in master was undefined. I assumed that a release hadn't been cut since that was merged.

Walking through git history and looking at when the `version.rb` file was tweaked, I think:
  * `v0.4.3` is accurate in the `CHANGELOG`
  * `v0.4.4` actually contains [this commit](https://github.com/discourse/prometheus_exporter/commit/b3faf90d3ff5ba6671a145c5525e38d4c694cf9d), which was not listed in the `CHANGELOG`
  * `v0.4.5` (latest version) contains what is currently listed in the `CHANGELOG` for `v0.4.4`, [this commit](https://github.com/discourse/prometheus_exporter/commit/32c3250855004fead1cc3c575e3b7729df241601)
  * What is currently listed for `v0.4.5` is not in any RubyGems release. I verified this by downloading the `.gem` from Rubygems, unpacking it, and looking at the source.

If you click on those commits I linked, Github is showing you what tags belong to. I think the git tags all jive with what is released the Rubygems; it's only the `CHANGELOG` that's incorrect. It's also a little confusing that there is code in the master `README` that does not exist in the latest RubyGems release.

I've updated the `CHANGELOG` with what I think is correct. Let me know if that makes sense. As far as updating the `README`, unless I made a new section documenting all the Metric classes, I'm not sure there's anything to update.

SamSaffron:
OK lets merge this in! Will do a new gem release today. 

Regarding the README ... I support a new section to document metrics if you are feeling adventurous!

PR #65, opened 2019-04-04
Add unicorn worker worker instrumentation

### LINKED ISSUE #64: Instrument unicorn server out of the box
If we had a server instrumentation for unicorn - which works almost the same as the puma instrumentation - would you be willing to consider that for merging? 

I'd have one "ready" in our code base, and we could prepare this for merging in if you'd approve of that feature.

Thanks already!

siebertm:
Fixes #64 

This uses `raindrops` to instrument unicorn workers.

Exposes:

* `ruby_unicorn_active_workers_total`
* `ruby_unicorn_request_backlog_total`
* `ruby_unicorn_workers_total`


SamSaffron:
Pattern wise, I think it is  better to advice placing the code in the collector process it is a cleaner spot for it and nice not to be interfering with the unicorn master process. Can you have a look at moving this? We do this for Discourse, raindrops should be fine to run in a diff process.

siebertm:
> Pattern wise, I think it is better to advice placing the code in the collector process it is a cleaner spot for it and nice not to be interfering with the unicorn master process.

Will do! 

Do you have any advice on how to get hold of the number of unicorn workers present? This make a nice "load" monitoring (e.g. when 80% of my workers are constantly loaded, I might want to do something)

Right now, I'm getting this off of the unicorn config which would not be possible anymore when I move that code out...

SamSaffron:
It really depends on your setup, but if you want something that is robust even to multi masters running in the same OS, you could pass the PID of the unicorn master to the collector on launch. Then you can just looks at the processes even with `ps aux` searching for the children of the master. 1 `ps aux` every 30 secs or  so.

siebertm:
@SamSaffron Here you go! Is that about what you were thinking of?

siebertm:
(The test failure is very likely a flapping test caused by unproper cleanup if `Base.default_prefix` in other tests... I can't find the "rerun" button on travis)

README.md
```diff
@@ -275,6 +277,18 @@ after_worker_boot do
 end
 ```
 
+### Unicorn process metrics
+
+In order to gather metrics from unicorn processes, we use `rainbows`, which exposes `Rainbows::Linux.tcp_socket_stats` to gather information about active workers and queued requests. To start monitoring your unicorn processes, you'll need to know both the path to unicorn PID file and the listen socket (`pid_file` and `listen` in your unicorn config file)
```
SamSaffron: Rainbows typo here I think :)

lib/prometheus_exporter/instrumentation/unicorn.rb
```diff
@@ -0,0 +1,68 @@
...
+
+      # find all processes whose parent is the unicorn master
+      # but we're actually only interested in the number of processes (= lines of output)
+      result = `ps --no-header -o pid --ppid #{pid}`
+      result.lines.count
+    end
+
+    def socket_stats
+      if @tcp
+        Raindrops::Linux.tcp_listener_stats(@socket)[@socket]
```
SamSaffron: This can take either an address or a socket per: https://www.rubydoc.info/gems/raindrops/0.13.0/Raindrops%2FLinux.tcp_listener_stats so we should clean up the params here a bit.

siebertm: Sorry, I dont really get what you mean here... :p

README.md
siebertm: Whoops, fixed

lib/prometheus_exporter/instrumentation/unicorn.rb
SamSaffron: sorry I mean this is legit: (in fact it is what we do in discourse) 

```
Raindrops::Linux.tcp_listener_stats(['127.0.0.1:3000'])['127.0.0.1:3000']
```

So maybe add an extra param to the collector so this is clear `--unicorn-address http://localhost:3000`

siebertm: `@socket` is that address... I'd rename that param to `listener_address` to be clear?

SamSaffron: yeah I guess that works ... was thinking maybe people want to be explicit if they use a `unicorn.sock` sock file here, but address works I guess.

SamSaffron:
This looks good,  merging it in and doing a release! Thanks

PR #70, opened 2019-04-19
Raindrops::Linux methods expect array

aiharak:
Hi,

I've faced a following error when using unicorn instrumentation.

```
2019-04-19 16:18:11 +0900 Starting prometheus exporter on port 9394
Prometheus Exporter Failed To Collect Unicorn Stats undefined method `map' for "/path/to/sock":String
Did you mean?  tap
```

I think that these methods expect array as an arg.

ref.
https://github.com/tmm1/raindrops/blob/1c18fd9c13f95fef6bcbdc0587d38886fa8e9064/lib/raindrops/linux.rb#L37
https://github.com/tmm1/raindrops/blob/1c18fd9c13f95fef6bcbdc0587d38886fa8e9064/ext/raindrops/linux_inet_diag.c#L574

Currently `@listener_address` is String, so I changed it to pass as an array.

### DIFF
--- a/lib/prometheus_exporter/instrumentation/unicorn.rb
+++ b/lib/prometheus_exporter/instrumentation/unicorn.rb
@@ -61,9 +61,9 @@ module PrometheusExporter::Instrumentation
 
     def listener_address_stats
       if @tcp
-        Raindrops::Linux.tcp_listener_stats(@listener_address)[@listener_address]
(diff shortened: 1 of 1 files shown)


SamSaffron:
@siebertm any feedback here? is this just relevant to socket connection?

siebertm:
TBH, we're not using unix sockets and this might have been slipped through. Definite :+1: from me!

SamSaffron:
cool, thanks @aiharak !

PR #71, opened 2019-04-25
Custom labels support in puma collector

gsmetal:
There are custom_labels in client, but puma collector does not support them when other collectors (hutch, sidekiq, etc) do.


SamSaffron:
Oh, nice fix! thanks!

PR #72, opened 2019-05-13
Handle stopping a nil worker_thread client

akofink:
`@worker_thread` in the code below can be nil, as evidenced by the nil check on line 126. Without this change, the following occurs when trying to stop a disconnected client:

```rb
> PrometheusExporter::Client.default.stop         
NoMethodError: undefined method `alive?' for nil:NilClass
```


SamSaffron:
looks good to me ... will do a quick release now

PR #76, opened 2019-06-07
jobs with retry = false should not count towards dead jobs counter

alexmorten:
We use the `sidekiq_dead_jobs_total` metric to alert us about jobs that ended up in the dead set.
Jobs with `retry: false` [don't end up in the dead set](https://github.com/mperham/sidekiq/wiki/Error-Handling#dead-set), but still call the `death_handler` and count towards the `sidekiq_dead_jobs_total` metric, leading to inconsistencies between the metric and what you can observe in Sidekiq.

tldr: I assume that jobs with `retry: false` should not count towards `sidekiq_dead_jobs_total`.

What do you think about this?


SamSaffron:
Yeah this change looks correct to me! thanks!

Issue #69, opened 2019-04-17
Expose a health check endpoint?


duanshiqiang:
Currently we are facing issues that the collector process stops working after running for some time (few days) and the /metrics response having "ruby_collector_working 0".

All of our workloads are running in kubernetes and we are running the collector process in a sidecar container. So if the collector web server (WEBrick process) can have a health check endpoint like '/healthz' then we can use it as the container livenessProbe so kubernetes can help us restart the container when the collector stops working.

SamSaffron:
This is a bit odd, we have not seen collectors die across our infrastructure at all. 

This metric is there to inform the central server that the collector has no metrics. My guess is that this would happen if you restarted the collector container and somehow your old containers were no longer able to talk to it. 

I think the remediation here is actually to restart the entire pod. 

Quick question first, do you have IPv6 stability across container destroy/create? Are you communicating via v4? If so do you have port stability?

duanshiqiang:
Hi @SamSaffron Thanks for your reply. We are using ipv4 only in our infrastructure.

We have subclassed `PrometheusExporter::Server::WebCollector` and running the collector with command like this:

`bundle exec prometheus_exporter -a lib/prometheus/web_with_tenant_collector.rb`

What we are observing is that the memory usage of the prometheus_exporter can go up even to over 1 GB which is really strange.

Also what you are implying in your reply is that you are running the prometheus_exporter as a central server? We are running it as a sidecar of each puma server (we are having over 90 pods) which I guess is not recommended. And I do find that since some metrics are of type `summary`, especially those percentiles metrcs, we are not able to write promql to recalculate it among all the server instances which is quite sad.

One thing we are afraid of running the prometheus_expoter as a central server is that there's a single point of failure and we found if the prometheus_expoter is down, our puma server will be flooded by error logs like:

`Prometheus Exporter, failed to send message Cannot assign requested address - connect(2) for "localhost" port 9394`

SamSaffron:
1GB in a collector is a big concern that would mean some list is ever growing, summary metrics do have a roof on them, but I guess if you keep adding metrics problems can arise. 

Can you look at one of your collectors say at 512MB and see how many metrics you have reported? then maybe do a ruby heap dump via rbtrace to see where the leak is. 

You can run as many collectors as you need but certainly there is a memory impact, that said it should be small in the big scheme. 

This is one of our collectors on a very busy server that has been running since March 27 (RSS is merely 36872)

```
[...]

PR #80, opened 2019-07-09
delete old metrics in puma collector

Fivell:
refs #69 #34

### DIFF
--- a/lib/prometheus_exporter/server/puma_collector.rb
+++ b/lib/prometheus_exporter/server/puma_collector.rb
@@ -2,6 +2,7 @@
 
 module PrometheusExporter::Server
   class PumaCollector < TypeCollector
+    MAX_PUMA_METRIC_AGE = 30
     PUMA_GAUGES = {
       workers_total: "Number of puma workers.",
       booted_workers_total: "Number of puma workers booted.",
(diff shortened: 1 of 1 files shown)


Fivell:
image

Fivell:
@SamSaffron  can't figure out why travis fails under 2.6 ruby , seems some technical problem

lib/prometheus_exporter/server/puma_collector.rb
```diff
@@ -47,6 +48,15 @@ def metrics
     end
 
     def collect(obj)
+      now = ::Process.clock_gettime(::Process::CLOCK_MONOTONIC)
+
+      obj["created_at"] = now
+
+      @puma_metrics.delete_if do |current|
+        (obj["pid"] == current["pid"] && obj["hostname"] == current["hostname"]) ||
```
eviltrout: I took a look at the puma collector and I don't think `pid` and `hostname` are included in the `obj`?

Fivell: @eviltrout you're right , seems we need to leave only current["created_at"] + MAX_PUMA_METRIC_AGE < now

eviltrout:
Thanks, I adapted this in e2d75a5aa84bc71e8bfcf07d8a406e1145ed9958 and credited you as a co-author.

PR #84, opened 2019-09-02
Guard socket closing just like socket opening

benny-burkhart:
This seems required in order to reliably stop metrics collection while logging the last metric just before stopping the process


SamSaffron:
ok, this looks like a reasonably safe change, thanks!

PR #86, opened 2019-09-04
Feature: allow finding registered metric by metric name, type, or help

FunnyHector:
When reusing the same client in different places to export the same
metric, there is no external interface to retrieve this metric after it's 
registered. Using `instance_variable_get` to reach into the instance 
variable `@metrics` is less preferable. Re-registering the metric will 
lead to an ever-growing array of `@metrics`.

This commit provides a way to get the metric by name, and optionally by
type and help, so that we can reuse it.

### DIFF
--- a/lib/prometheus_exporter/client.rb
+++ b/lib/prometheus_exporter/client.rb
@@ -7,6 +7,8 @@ module PrometheusExporter
 
   class Client
     class RemoteMetric
+      attr_reader :name, :type, :help
+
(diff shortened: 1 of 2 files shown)


SamSaffron:
This looks good to me, mind adding a test for it?

FunnyHector:
> This looks good to me, mind adding a test for it?

Yep, happy to :)

SamSaffron:
Looks good to me! thanks!

PR #87, opened 2019-09-25
Explicitly require Client class

sgerrand:
💁 Many of the instrumentation implementations reference `PrometheusExporter::Client` without explicitly `require`ing that class first. Adding a `require` for that class here means they can be included as one file reference by consumers.

### DIFF
--- a/lib/prometheus_exporter/instrumentation.rb
+++ b/lib/prometheus_exporter/instrumentation.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require_relative "client"
 require_relative "instrumentation/process"
 require_relative "instrumentation/method_profiler"
 require_relative "instrumentation/sidekiq"


sgerrand:
For example, using the current documentation for the `DelayedJob` exporter with
```
unless Rails.env == "test"
  PrometheusExporter::Instrumentation::DelayedJob.register_plugin
end
```
in `config/initializers/delayed_job.rb` results in
```
web_1                  | bundler: failed to load command: unicorn (/app/vendor/bundle/ruby/2.5.0/bin/unicorn)
web_1                  | NameError: uninitialized constant PrometheusExporter::Client
web_1                  |   /app/vendor/bundle/ruby/2.5.0/gems/prometheus_exporter-0.4.13/lib/prometheus_exporter/instrumentation/delayed_job.rb:26:in `initialize'
web_1                  |   /app/vendor/bundle/ruby/2.5.0/gems/prometheus_exporter-0.4.13/lib/prometheus_exporter/instrumentation/delayed_job.rb:9:in `new'
web_1                  |   /app/vendor/bundle/ruby/2.5.0/gems/prometheus_exporter-0.4.13/lib/prometheus_exporter/instrumentation/delayed_job.rb:9:in `register_plugin'
web_1                  |   /app/config/initializers/delayed_job.rb:5:in `<top (required)>'
```
💥

SamSaffron:
Sure, this feels pretty safe to me, thanks!

PR #90, opened 2019-10-16
Allow custom labels for Unicorn metrics

LS80:
The unicorn and process collectors do not respect the `custom_labels` set on the client.

I need to be able to add a custom label (e.g. host name) to aggregate the unicorn metrics across multiple instances.

With a `hostname` custom label I now get
```
ruby_unicorn_workers_total{hostname="bb4c06c24e81"} 2
ruby_unicorn_workers_total{hostname="d081f77622c7"} 2
```

instead of
```
ruby_unicorn_active_workers_total 2
```

Also makes the client optional for the unicorn instrumentation, so I can make this change to my unicorn config file

```diff
port_http = ENV.fetch('PORT_HTTP')
pid_file = '/tmp/unicorn.pid'

PrometheusExporter::Instrumentation::Unicorn.start(
  pid_file: pid_file,
  listener_address: "0.0.0.0:#{port_http}",
- client: nil,
)
```

### DIFF
--- a/lib/prometheus_exporter/instrumentation/unicorn.rb
+++ b/lib/prometheus_exporter/instrumentation/unicorn.rb
@@ -9,7 +9,7 @@ end
 module PrometheusExporter::Instrumentation
   # collects stats from unicorn
   class Unicorn
-    def self.start(pid_file:, listener_address:, client:, frequency: 30)
+    def self.start(pid_file:, listener_address:, client: nil, frequency: 30)
       unicorn_collector = new(pid_file: pid_file, listener_address: listener_address)
       client ||= PrometheusExporter::Client.default
       Thread.new do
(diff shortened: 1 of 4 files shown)


lib/prometheus_exporter/instrumentation/unicorn.rb
```diff
@@ -26,15 +26,17 @@ def self.start(pid_file:, listener_address:, client:, frequency: 30)
...
       @pid_file = pid_file
       @listener_address = listener_address
       @tcp = listener_address =~ /\A.+:\d+\z/
+      @metric_labels = metric_labels
     end
 
     def collect
       metric = {}
       metric[:type] = 'unicorn'
+      metric[:metric_labels] = @metric_labels
```
SamSaffron: minor, but lets simply not send this over to the other side if it is empty, just allow it to be nil, if nil don't pass it. then in the collector if it is nil you do the old observe

LS80: I realised we can just use the `custom_labels` set on the client. I've updated the PR accordingly.

SamSaffron:
Thanks! let's add a spec to cover this so the behavior is properly documented. Then I am good merge.

LS80:
I've added the test case 👍

SamSaffron:
Thanks heaps! merging this in.

PR #91, opened 2019-10-24
Histogram and summary options on remote metrics 

hoffoo:
this PR rebase and applies fixes from here https://github.com/discourse/prometheus_exporter/pull/46

@SamSaffron Thanks!

### DIFF
--- a/lib/prometheus_exporter/client.rb
+++ b/lib/prometheus_exporter/client.rb
@@ -9,12 +9,12 @@ module PrometheusExporter
     class RemoteMetric
       attr_reader :name, :type, :help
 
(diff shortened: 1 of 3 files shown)


lib/prometheus_exporter/client.rb
```diff
@@ -9,18 +9,20 @@ class Client
     class RemoteMetric
       attr_reader :name, :type, :help
 
-      def initialize(name:, help:, type:, client:)
+      def initialize(name:, help:, type:, client:, opts: {})
```
SamSaffron: can't we just flow `opts: nil` all the way through and simply not send it over if it is nil?

hoffoo:
closed in favor of #46

PR #98, opened 2020-01-09
FEATURE: Support #to_h on all metric types to fetch label value set for a metric

techAPJ:


SamSaffron:
I am not a fan of the name `keys` for this method. Cause 

```
{a: 1, b: 2}.keys
[:a, :b]
```

`values` that prom client uses has the exact same surprise going for it. 

Instead... I think the most Ruby-ish thing we can to is probably define `to_h` (rename keys, to_h)

Also make sure you never leak out the internal store, you should do a `.dup` in Gauge and Counter.

lib/prometheus_exporter/metric/counter.rb
```diff
@@ -23,6 +23,10 @@ def metric_text
       end.join("\n")
     end
 
+    def to_h
+      @data
```
SamSaffron: lets add a .dup here.

lib/prometheus_exporter/metric/gauge.rb
```diff
@@ -23,6 +23,10 @@ def reset!
       @data = {}
     end
 
+    def to_h
+      @data
```
SamSaffron: .dup

SamSaffron:
feel free to merge once the dups are in.

PR #100, opened 2020-01-20
adding socket creation exception handling

rajkumarkandasami:
when no prometheus exporter collector running (in rails environment), exception got thrown in rails console. Which is caught and handled.


SamSaffron:
I am not sure about this, the buffer will just keep on growing and growing and nobody will know a problem is going on.

rajkumarkandasami:
Hi Sam, 
I will explain the PR,
My setup has two containers
1. rails app instrumented with prometheus exporter
2. prometheus exporter collector process (which we typically ran by "bundle exec prometheus_exporter") in a separate container

both containers sharing network namespace, so the app container can talk with collector process through 9394 port. 

Case 1 :
Let's say the collector container not running, It tries to create a socket, 
`@socket = TCPSocket.new @host, @port`
Which It cannot.

Case 2:
If 9394 port is not exposed in collector host/container,

In both the cases client sends error message in rails console as
`irb(main):001:0> Prometheus Exporter, failed to send message Connection timed out - connect(2) for "localhost" port 9394
`
which should be rescued.

PR #102, opened 2020-01-21
fix unicorn example

and0x000:
[Unicorn::Configurator complains with ArgumentError](https://github.com/honeybadger-io/unicorn/blob/master/lib/unicorn/configurator.rb#L622) unless block parameters are passed


SamSaffron:
looks good, thanks!

Issue #101, opened 2020-01-20
allow specification of exporter bind address


and0x000:
I think it would be a good idea, to allow the specification of an address to which [PrometheusExporter::WebServer](https://github.com/discourse/prometheus_exporter/blob/master/lib/prometheus_exporter/server/web_server.rb#L39) should bind to.

See [WEBrick::HTTPServer -> :BindAddress](https://ruby-doc.org/stdlib-2.7.0/libdoc/webrick/rdoc/WEBrick/HTTPServer.html)

Going a step further, binding to `localhost` only should probably become the default behavior. This is the way, several other pieces of software do it to avoid unintended insecure configurations.

SamSaffron:
I am fine for starters for this to be configurable and documented, care to post a PR?

PR #103, opened 2020-01-21
add option for binding exporter to ip

and0x000:
binds to localhost only by default, preventing unintended exposure of this service

closes https://github.com/discourse/prometheus_exporter/issues/101

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ require 'prometheus_exporter/instrumentation'
 
 # bind is the address, on which the webserver will listen
 # port is the port that will provide the /metrics route
-server = PrometheusExporter::Server::WebServer.new bind: localhost, port: 12345
+server = PrometheusExporter::Server::WebServer.new bind: 'localhost', port: 12345
 server.start
 
 # wire up a default local client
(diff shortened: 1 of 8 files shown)


and0x000:
Feel free to edit the PR. Setting `PrometheusExporter::DEFAULT_BIND_ADDRESS` to '*' instead of 'localhost' should keep the current behavior, where it binds to all addresses. However, I honestly believe, defaulting to 'localhost' would be best practice.

SamSaffron:
OK I can accept the behavior change, it is breaking so we will need to be careful here.

Can you amend the README and CHANGELOG as well in light of this?

README.md
```diff
@@ -62,8 +62,9 @@ require 'prometheus_exporter/server'
 require 'prometheus_exporter/client'
 require 'prometheus_exporter/instrumentation'
 
+# bind is the address, on which the webserver will listen
 # port is the port that will provide the /metrics route
-server = PrometheusExporter::Server::WebServer.new port: 12345
+server = PrometheusExporter::Server::WebServer.new bind: localhost, port: 12345
```
SamSaffron:
```suggestion
server = PrometheusExporter::Server::WebServer.new bind: 'localhost', port: 12345
```

SamSaffron:
OK lets give this a shot! Thanks!

PR #104, opened 2020-01-23
Adding Shoryuken metrics

rajkumarkandasami:


SamSaffron:
looks good, can we add a test or two here just to at least make sure the collector is good?

rajkumarkandasami:
> looks good, can we add a test or two here just to at least make sure the collector is good?

Hi Sam, a basic metric collection test is added

SamSaffron:
Looks good to me! thanks!

PR #108, opened 2020-02-24
FEATURE: Allow configuring the default client's host and port via environment variables

siebertm:
In my docker-compose setup, I have to run prometheus_exporter server and client in different containers on the same network. By default, all the integrations use `PrometheusExporter::Client.default` which uses `localhost`, all the time

This MR defaults host and port to also look at the environment variable `PROMETHEUS_EXPORTER_HOST` and `PROMETHEUS_EXPORTER_PORT` to allow easy override using the environment configuration.


SamSaffron:
Yeah I think this makes sense it does make using the client a bit easier.

siebertm:
Great, thanks a lot! Do you have any plans for releasing this soon(ish)?

SamSaffron:
Just pushed 0.5.1 :)

Issue #110, opened 2020-02-27
Setting default label in bin/prometheus_exporter


weewey:
Currently in the `PrometheusExporter::Metric::Base`, default_label is not being set.

I believe we should be able to set it in `bin/prometheus_exporter` similar to how we can set the default_prefix.

- Adding an additional argument in the OptionsParser
- Passing `default_label` to the Runner
- Setting `default_label` at the `start` method of Runner

weewey:
Sent a PR: https://github.com/discourse/prometheus_exporter/pull/111

PR #111, opened 2020-02-28
Set default labels at bin/prometheus_exporter

weewey:
Referencing to this issue: https://github.com/discourse/prometheus_exporter/issues/110

### DIFF
--- a/bin/prometheus_exporter
+++ b/bin/prometheus_exporter
@@ -2,6 +2,7 @@
 # frozen_string_literal: true
 
 require 'optparse'
+require 'json'
 
 require_relative "./../lib/prometheus_exporter"
 require_relative "./../lib/prometheus_exporter/server"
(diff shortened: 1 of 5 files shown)


bin/prometheus_exporter
```diff
@@ -34,6 +35,9 @@ def run
     opt.on('--prefix METRIC_PREFIX', "Prefix to apply to all metrics (default: #{PrometheusExporter::DEFAULT_PREFIX})") do |o|
       options[:prefix] = o.to_s
     end
+    opt.on('--label METRIC_LABEL', "Label to apply to all metrics (default: #{PrometheusExporter::DEFAULT_LABEL})") do |o|
+      options[:label] = JSON.parse(o.to_s)
```
SamSaffron: Do we need a full JSON.parse here?

weewey: `bin/prometheus_exporter --label '{"environment": "integration"}'`

Because a label requires a hash, that's why it requires a JSON.parse. Unless we breakdown the key and value in different arguments. I didn't choose that method because it will be harder to specify multiple labels.

`bin/prometheus_exporter --label '{"environment": "integration", "test": "valueA"}'`
VS
`bin/prometheus_exporter --label-key environment,test --label-value integration,valueA`

weewey: @SamSaffron any feedback?

SamSaffron:
I think this looks good, lets merge it... the arg via JSON is a bit surprising, can you add a change to the README.md that explains how to use this with an example. 

Once we have that lets merge.

weewey:
Just done. Cheers! :)
@SamSaffron

SamSaffron:
Looks good to me ! Thanks

PR #112, opened 2020-03-13
added sidekiq job queue time metric and queue name tag in default labels

rajkumarkandasami:


test/server/collector_test.rb
```diff
@@ -48,11 +47,11 @@ def test_register_metric
     collector.register_metric(metric)
     metric.observe(77)
     metric.observe(2, red: "alert")
-    text = <<~TXT
-      # HELP amazing amount of amazing
-      # TYPE amazing gauge
-      amazing 77
-      amazing{red="alert"} 2
+    text = <<-TXT
```
SamSaffron: Why this change?

lib/prometheus_exporter/server/sidekiq_collector.rb
```diff
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
```
SamSaffron: This?

lib/prometheus_exporter/server/sidekiq_collector.rb
```diff
@@ -43,13 +43,17 @@ def ensure_sidekiq_metrics
       if !@sidekiq_jobs_total
 
         @sidekiq_job_duration_seconds =
-        PrometheusExporter::Metric::Counter.new(
+        PrometheusExporter::Metric::Gauge.new(
```
SamSaffron: This really is a counter...

Issue #88, opened 2019-10-07
tracking sidekiq queue size and latecy


jwhitcraft:
Hello,

would it be possible to have the sidekiq metrics extended to also include the queue size and latency for any given queue?

Thanks,

SamSaffron:
I am fine with some examples being added to the readme, technically this is
not complicated to add.

PR #114, opened 2020-05-29
Add collector and instrumentation for Sidekiq queues

mrsimo:
Hi :wave:!

We're using this to track the backlog and latency for Sidekiq queues. Besides being very useful metrics, we believe latency is the metric to use to monitor Sidekiq's health. All our alerts are based on it. I saw there was an issue for it already: https://github.com/discourse/prometheus_exporter/issues/88

I think it would be great if `prometheus_exporter` had this, it would encourage more people to monitor these very useful things. 

I saw there's no tests for the Sidekiq things... it would be a bit tricky I guess. I'm very open to suggestions about that (or anything else about the PR, really).

Thanks a lot for your time!


SamSaffron:
Yeah I very much support this, we use a similar metric at Discourse.

I would love some tests, I think we can test the collector standalone at least? 

The instrumenter requires mocking so it is a bit of a non test.

mrsimo:
Cool!

> I would love some tests, I think we can test the collector standalone at least?

I've pushed some tests, had to refactor the collector to look even more like the others and not instantiate the gauges until the first metrics are collected.

> The instrumenter requires mocking so it is a bit of a non test.

Yeah... I don't think we can get much value testing that class without a super complex setup.

mrsimo:
I noticed there's some rubocop configuration that isn't working properly... which is also breaking Travis.  I can try to help with that tomorrow, but I'm not familiar with the rubocop setup here.

SamSaffron:
It is looking good to me! we can sort out the rubocop stuff, I have a feeling something needs updating in the repo!

Issue #115, opened 2020-06-10
Documentation, list of metric series


wolfgangrittner:
Hey there, we're using prometheus-exporter and it's great so far! Thank you for that!
What we are missing, and what I could not really find was a list of metrics series prometheus-exporter itself and all the various integrations (sidekiq, puma etc.) create and report to. Skimming through the code works, but gets tedious.
Would you be open for a PR that tries to add an extensive list of series created and used by the various parts of prometheus-exporter to the README? Or would you want that to go somewhere else? Or does that exist already after all and I totally missed it? 😬

SamSaffron:
> Would you be open for a PR that tries to add an extensive list of series created and used by the various parts of prometheus-exporter to the README?

Absolutely! That would be delightful.

wolfgangrittner:
Guess this can be closed?

SamSaffron:
sure can, thanks!!

PR #116, opened 2020-06-14
Add list of metrics series to readme

wolfgangrittner:
Hey there, this adds some info on all 🤞 metrics series produced by the various integrations to the README - as discussed #115.

I was torn between adding one big table that lists everything somewhere at the end of the README and with what I eventually came up with, which is adding a small overview to each section just listing the metrics produced by a single integration.
Let me know if that works for you. Any feedback is very much appreciated!

### DIFF
--- a/README.md
+++ b/README.md
@@ -482,6 +482,20 @@ after_worker_boot do
 end
 ```
 
+#### Metrics collected by Puma Instrumentation
+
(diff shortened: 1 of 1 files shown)


SamSaffron:
Wow ... thanks so much :heart:

PR #127, opened 2020-09-10
Augment default metrics

etnt93:
We'd like to see the ability to add custom labels to default metrics. For example, adding 
`labels['platform'] = env['HTTP_X_PLATFORM']` to `http_requests_total`. 

Also added 'queue' as a label for default sidekiq metrics in order to view metrics on sidekiq jobs at the 'queue' level.


SamSaffron:
This is conflicting now, can you have a look?

etnt93:
Resolved merge conflicts!

lib/prometheus_exporter/middleware.rb
```diff
@@ -44,14 +44,25 @@ def call(env)
...
+    if labels.present? && labels.length > 0
+      obj = obj.merge(custom_labels: labels)
+    end
+
+    @client.send_json(obj)
+  end
+
+  # allows subclasses to add custom labels based on env
+  def custom_labels(env)
+    {}
```
SamSaffron, line 65: let's return nil here, then we can avoid all this merging in the client when labels are not configured.

Also present? is going to be an active support thing I think

etnt93: Returns `nil` and removed `present?` in the latest commit.

lib/prometheus_exporter/client.rb
```diff
@@ -107,7 +107,7 @@ def find_registered_metric(name, type: nil, help: nil)
     end
 
     def send_json(obj)
-      payload = @custom_labels.nil? ? obj : obj.merge(custom_labels: @custom_labels)
+      payload = @custom_labels.nil? ? obj : obj.merge(custom_labels: @custom_labels.merge(obj[:custom_labels] || {}))
```
SamSaffron, line 110: minor but this is perf sensitive. No need to merge here if obj[:custom_labels] is nil. 

```
payload = 
  if @custom_labels
     if obj[:custom_labels]
         ...
     else
        ...
     end 
  else
    obj
  end
  
```

lib/prometheus_exporter/middleware.rb
```diff
@@ -44,14 +44,25 @@ def call(env)
...
+    }
+    labels = custom_labels(env)
+    if labels
+      obj = obj.merge(custom_labels: labels)
+    end
+
+    @client.send_json(obj)
+  end
+
+  # allows subclasses to add custom labels based on env
```
SamSaffron, line 63: very important ... we are making an api change here so we must add something to the README about how to consume this.

SamSaffron:
cool let's try this out... also our CI is playing up, got to get that sorted.

PR #128, opened 2020-09-10
Add HTTP status to http_duration_seconds

dmcaulay:
It'd be great to compare counts and duration by HTTP status code.

### DIFF
--- a/lib/prometheus_exporter/server/web_collector.rb
+++ b/lib/prometheus_exporter/server/web_collector.rb
@@ -62,11 +62,16 @@ module PrometheusExporter::Server
       }
       custom_labels = obj['custom_labels']
       labels = custom_labels.nil? ? default_labels : default_labels.merge(custom_labels)
+      http_labels = labels.merge(status: obj["status"])
 
-      @http_requests_total.observe(1, labels.merge(status: obj["status"]))
+      @http_requests_total.observe(1, http_labels)
 
       if timings = obj["timings"]
-        @http_duration_seconds.observe(timings["total_duration"], labels)
+        if obj['http_labels_on_duration']
+          @http_duration_seconds.observe(timings["total_duration"], http_labels)
+        else
+          @http_duration_seconds.observe(timings["total_duration"], labels)
+        end
         if redis = timings["redis"]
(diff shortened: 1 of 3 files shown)


SamSaffron:
lets add a test for this, just to specify what stats look like

dmcaulay:
> lets add a test for this, just to specify what stats look like

sounds good. i added assertions for `http_duration_seconds`.

SamSaffron:
oh now I see what is happening, lets make this an opt in thing. 

I worry that this explodes the amount of metrics, in general we don't ever look at this level of fidelity at discourse. 

You will end up getting summaries for 500s 401s 403s 301s 302s ... the amount of metrics and counting is super high.

dmcaulay:
> oh now I see what is happening, lets make this an opt in thing.
> 
> I worry that this explodes the amount of metrics, in general we don't ever look at this level of fidelity at discourse.
> 
> You will end up getting summaries for 500s 401s 403s 301s 302s ... the amount of metrics and counting is super high.

Makes sense. I wasn't sure on the best approach to making it optional. Let me know what you think.

lib/prometheus_exporter/server/web_collector.rb
```diff
@@ -62,11 +62,16 @@ def observe(obj)
...
       custom_labels = obj['custom_labels']
       labels = custom_labels.nil? ? default_labels : default_labels.merge(custom_labels)
+      http_labels = labels.merge(status: obj["status"])
 
-      @http_requests_total.observe(1, labels.merge(status: obj["status"]))
+      @http_requests_total.observe(1, http_labels)
 
       if timings = obj["timings"]
-        @http_duration_seconds.observe(timings["total_duration"], labels)
+        if obj['http_labels_on_duration']
```
SamSaffron, line 70: This kind of feels like the responsibility of the collector, no need to send this extra information downstream. Perhaps we allow specifying config for collector initialization?

dmcaulay:
i think your point about fidelity makes a lot of sense. the most i'd actually want is 1xx, 2xx, 3xx, 4xx, 5xx, but then it deviates from counters. maybe best to hold off on this one. we can easily bring it back if we want.

PR #130, opened 2020-09-10
Parse Sidekiq Delayed class names

dmcaulay:
Adjusting the job name for Sidekiq Delayed classes gives users more insight into which Mailer, Model or Class methods are contributing to high counts or durations. For example, `UserMailer.delay.welcome(user.id)` would now have the job name of `UserMailer#welcome` instead of `Sidekiq::Extensions::DelayedMailer`.


SamSaffron:
Looks good to me!

PR #132, opened 2020-10-05
Use unfrozen version of Symbol#to_s

fauno:
Makes it compatible with [Shopify/symbol-fstring](https://github.com/Shopify/symbol-fstring)

Otherwise it throws `Prometheus Exporter Failed To Collect Process Stats can't modify frozen String: "database"` periodically.


SamSaffron:
Oh, good catch! thanks.

PR #135, opened 2020-10-23
Fix Sidekiq issues

dmcaulay:
This PR contains the following changes:

* Fixes broken tests after the introduction of sidekiq queue to stats (823710f)
* Fixes fallback to `class_name` in the `get_delayed_name` helper
* Adds additional logic to `get_delayed_name` so it handles [ActiveRecord](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/extensions/active_record.rb) properly- it turns out the entire object is serialized


SamSaffron:
I see, let's give this a shot! thanks.

PR #137, opened 2020-11-10
Add support for HTTPAuth::BasicAuth in prometheus_exporter

dyoung522:
Adds support for basic authentication in the prometheus_exporter CLI application.

* When given a `htpasswd` filename via the `--auth` command line argument, the server will use this file to authenticate any requests to the `/metrics` endpoint.
* Optionally, a custom REALM may be provided via the `--realm` argument. The default REALM is "Prometheus Exporter"

_Note: the htpasswd file must be created beforehand and, due to limitations in WEBrick, only crypt is supported._

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+0.6.0 - 10-11-2020
+
+- FEATURE: add support for basic-auth in the prometheus_exporter web server
+
 0.5.3 - 29-07-2020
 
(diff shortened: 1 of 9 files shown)


SamSaffron:
A very nice PR! thank you. I love that you cleaned up stuff as you went. 

I think the only big missing thing here is documentation in the README, can you add that?

SamSaffron:
Thanks! failure looks unrelated, change looks good to me! thanks heaps.

Issue #138, opened 2020-11-11
Delete routine is not specific enough for multiple pools in the same pid


vicomte:
For instance I have my standard connection opened by rails, but I also have some models in a different database that are opened by an establish_connection call. They are both sending metrics correctly to the prometheus_exporter, but the second pool metric immediately deletes the first since the key can't distinguish them. The delete is making comparison on  :pid, and :hostname. In my case adding a comparison on :metric_labels["pool_name"] is enough to fix this, but not sure if this is adequate for others. Perhaps If :metric_labels["pool_name"] exists and has data it could be added to the comparison on the fly.

PR #139, opened 2020-11-11
Suggestion to make deletion more granular.

vicomte:
From #138


lib/prometheus_exporter/server/active_record_collector.rb
```diff
@@ -47,11 +47,25 @@ def collect(obj)
       obj["created_at"] = now
 
       @active_record_metrics.delete_if do |current|
-        (obj["pid"] == current["pid"] && obj["hostname"] == current["hostname"]) ||
-          (current["created_at"] + MAX_ACTIVERECORD_METRIC_AGE < now)
+        obj_compare(obj,current, [:pool_name]) || (current["created_at"] + MAX_ACTIVERECORD_METRIC_AGE < now)
```
SamSaffron, line 50: if all we need is to add a check for pool_name do we need to iterate through all the keys and add the general purpose function? 

My pref here is to avoid over generalizing here until we need to.

vicomte: No problem. I didn't know if other's setups would make this entry unique with another key or not as I am somewhat focused atm.

PR #144, opened 2020-12-03
Add ActiveRecord 6.1 support

kvokka:
ActiveRecord 6.1.rc1 changed the syntax of the connection pool. 

This PR add the support of the new AR syntax, keeping AR 6.0 syntax still working.

Also it fix CI for Ruby 2.5, which apparently failed.


SamSaffron:
cool, can you add a CI workflow to test this? I guess we add an integration test that tests the metric can be collected and then run CI with both rails 6 and 6.1 ?

kvokka:
@SamSaffron Had to add a pluck for rubygems to support previous versions of Ruby, cos otherwise CI drove mad with provided gemfile option. Will be happy to apply simpler solution, if u have it.

Since we do not need rails, i install only active_record. This gem require AR >= 6.0 (which was before this PR and was implicit).

SamSaffron:
Looks good to me! thanks @kvokka !

PR #149, opened 2020-12-28
Add dependency on webrick

ball-hayden:
As of Ruby 3.0.0, Webrick is no longer bundled with Ruby (https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/)

Adds an explicit dependency on Webrick


SamSaffron:
Agree, we got to do this, technically we could make it only a dep for 3.0 and up, but wrestling that into a gemspec is impossible

PR #150, opened 2020-12-29
FEATURE: Support Amazon Application Load Balancer request tracing header

booleanbetrayal:
Thanks for this great Gem!

This PR implements support for the Amazon Application Load Balancer [request tracing headers](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-request-tracing.html) to support the Rails integration's [request queue timing](https://github.com/discourse/prometheus_exporter#instrumenting-request-queueing-time) metrics.

It will continue to favor `X-Request-Start` or `X-Queue-Start` headers but will fall back on the `Root` stamping of the `X-Amzn-Trace-Id` header, if present. The timestamp in this header is the number of seconds since epoch in hex.

### DIFF
--- a/README.md
+++ b/README.md
@@ -505,7 +505,7 @@ Request Queueing is defined as the time it takes for a request to reach your app
 
 As this metric starts before `prometheus_exporter` can handle the request, you must add a specific HTTP header as early in your infrastructure as possible (we recommend your load balancer or reverse proxy).
 
-Configure your HTTP server / load balancer to add a header `X-Request-Start: t=<MSEC>` when passing the request upstream. For more information, please consult your software manual.
+The Amazon Application Load Balancer [request tracing header](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-request-tracing.html) is natively supported. If you are using another upstream entrypoint, you may configure your HTTP server / load balancer to add a header `X-Request-Start: t=<MSEC>` when passing the request upstream. For more information, please consult your software manual.
 
 Hint: we aim to be API-compatible with the big APM solutions, so if you've got requests queueing time configured for them, it should be expected to also work with `prometheus_exporter`.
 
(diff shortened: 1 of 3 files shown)


SamSaffron:
Thanks, we a test here (and to confirm we have tests for apache / nginx), I can also see us doing a bit of micro optimising on this

booleanbetrayal:
Hi @SamSaffron - Found the specs and will get a backfill pushed. What micro-optimizations were you thinking?

SamSaffron:
was thinking maybe of avoiding the regex gsub, it is tiny perf wise ... 

Also probably worth making this code a bit more consistent and deliberate 

```
value = env['HTTP_X_REQUEST_START']
if value && value != '' 
   ...
end

value = env['HTTP_X_QUEUE_START']
if value && value != '' 
   ...
end

value = env['HTTP_X_AMZN_TRACE_ID']
if value && value != '' 
   ...
end
```

SamSaffron:
also I wonder if we get people to opt for the type of tracing they want (I want Amazon ... I  want NGINX ... etc) so we don't have to keep fishing in headers we don't care about.

booleanbetrayal:
@SamSaffron  -

I couldn't find a lot of firm documentation on `X-Queue-Start` and `X-Request-Start` header standards and documentation from sites like New Relic reference them as interchangeable. All that is to say, I didn't feel comfortable mucking with any of that logic. I did however get the specs backfilled and optimized around a string literal `split` for a decent benchmarking improvement:

```
now = '1234567890'
header = "Root=1-#{now.to_i.to_s(16)}-abc123"

def get_match_value(value)
  value&.match(/Root=.+\-([0-9a-f]+)-/)&.captures&.last&.to_s&.to_i(16)
end

def get_split_value(value)
  value&.split('Root=')&.last&.split('-')&.fetch(1)&.to_i(16)
end

require 'benchmark'

n = 1_000_000
Benchmark.bmbm do |x|
  x.report('match') { for i in 1..n; get_match_value(header); end }
  x.report('split') { for i in 1..n; get_split_value(header); end }
end

----

-------------------------------- total: 2.223692sec

            user     system      total        real
match   1.350277   0.004962   1.355239 (  1.358576)
split   0.842162   0.002418   0.844580 (  0.846389)
```

Did you want to make any cosmetic / additional optimization changes you might be thinking about, or is this suitable for merge as is?

SamSaffron:
I think we are good, thanks for the PR

SamSaffron:
HTTP_X_AMZN_TRACE_ID can be used to track delays between load balancer and app

SamSaffron:
Also, hoping for a better 2021! Take care.

booleanbetrayal:
Thanks Sam - You too!

PR #161, opened 2021-02-04
Fix active record support

### LINKED ISSUE #160: Fix ActiveRecord instrumentation
Got in the logs

```
Prometheus Exporter Failed To Collect Process Stats undefined method `spec' for #<ActiveRecord::ConnectionAdapters::ConnectionPool:0x000055b5bf49d1b0>
Did you mean?  inspect
```

with rails 6.1.1

kvokka:
It's much better to rely on gem version than on `respond_to?`, cos in my case i faced the situation when AR got the method dynamically from other gem and it leads to 💥 

This should fix it

Close #160

### DIFF
--- a/Appraisals
+++ b/Appraisals
@@ -6,5 +6,5 @@ appraise "ar-60" do
 end
 
 appraise "ar-61" do
-  gem "activerecord", "~> 6.1.0.rc2"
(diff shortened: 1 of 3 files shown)


SamSaffron:
looks good, let's try it out!

Issue #49, opened 2018-12-13
Change prefix on custom metrics


blackjid:
It is possible to change the prefix for a specific metric??

I would like to stay with the standard `ruby_` prefix for the built-in metrics, but change it for my custom ones.

SamSaffron:
I think we are going to need some sort of PR to support an override here.

SamSaffron:
I think we got to clean up the naming on all the process metrics, `heap_free_slots` is not a good name, it should be `ruby_heap_free_slots` and I think it should be prefix immune if we wish, so we need an extra flag to make it prefix immune. 

Unfortunately this will be a bit of a breaking change, but it is worth it for consistency sake.

thomasbalsloev:
We would welcome this feature as it seems that is the only way to get rid of these warnings in Google Cloud logging system: https://github.com/GoogleCloudPlatform/k8s-stackdriver/issues/202

But we can't add a custom metric with the required name due to the global nature of the prefix.

Best regards, Thomas.

SamSaffron:
not against this in principle, but let's start with API first

What API are your proposing?

thomasbalsloev:
Hi @SamSaffron 

Thanks for getting back on this.

I can image a couple of things to choose from:

1. On individual custom registered metrics.
client = PrometheusExporter::Client.default
gauge = client.register(:gauge, "awesome", "amount of awesome", prefix: 'custom_')

2. Global option for prefix on custom metrics.
```
Usage: prometheus_exporter [options]
    -p, --port INTEGER               Port exporter should listen on (default: 9394)
    -b, --bind STRING                IP address exporter should listen on (default: localhost)
    -t, --timeout INTEGER            Timeout in seconds for metrics endpoint (default: 2)
        --prefix METRIC_PREFIX       Prefix to apply to all metrics (default: ruby_)
     **--custom-metric-prefix STRING  Prefix to apply to all custom metrics (default: ruby_)**
```

BR, Thomas.

SamSaffron:
I like the idea of adding a `prefix: ...` to the register API, is this is PR you could work on?

thomasbalsloev:
@SamSaffron  

Hi Sam. 

Sure, I may not get the time until after Easter though, but I'll give it a shot :)

BR, Thomas.

PR #163, opened 2021-03-26
Add possibility for setting prefix on custom metrics, solves #49

thomasbalsloev:
Hi @SamSaffron 

Added possibility to set prefix on custom metrics. 
Let me hear what you think of the code enclosed.

BR, Thomas.


nevesenin:
Would also like to have this feature.

SamSaffron:
Yeah it is a nice change @thomasbalsloev sorry I left it so long, can you also add a note to README here that teaches about usage?

PR #168, opened 2021-04-27
Add source location for MethodProfiler patches

haines:
I ran into [an issue with the latest version of the New Relic gem](https://github.com/newrelic/newrelic-ruby-agent/issues/659) which was caused by a bad interaction with Prometheus Exporter's method profiling on the Redis client.

It was hard to debug because I found that the source location for the `Redis::Client#call` was `(eval):3` and I had no idea where the metaprogramming that was redefining the method was happening.

This PR adds the file and line number to the `class_eval` call, so that the patched method has a useful source location.


SamSaffron:
Nice fix!

PR #170, opened 2021-05-07
Add support for Resque

clupprich:
I'm in the need of Prometheus metrics for Resque, so I thought I could give this a try.

Right now I'm using [`Resque.info`](https://github.com/resque/resque/blob/7f6b88404dd18698f6f4023e18fdc1ae8318e5e5/lib/resque.rb#L561-L573) only, but I wanna extend this to fetch metrics per queue, not just totals. I'll update this PR with those changes in the next few days.

Happy about any feedback for the existing solution - I basically just copied the Puma integration, don't know if that's a good standard.


SamSaffron:
Looks nice and tidy! Are you happy to maintain this collector long term in case stuff pops up? 

Let me know when you feel it is ready to merge

clupprich:
Yes, happy to keep working on this and be the go-to person for issues that might arise!

I'll let you know once I'm happy with the current state.

clupprich:
@SamSaffron This is ready for review now. I did some small cleanups (harmonised variable naming with the Sidekiq collector). I still wanna do metrics per queue, but I'll followup with a separate PR for that once this is merged.

lib/prometheus_exporter/server.rb
```diff
@@ -16,3 +16,4 @@
 require_relative "server/unicorn_collector"
 require_relative "server/active_record_collector"
 require_relative "server/shoryuken_collector"
+require_relative "server/resque_collector"
```
SamSaffron, line 19: we should do a separate PR, but I wonder now that we have so many of these if we should just switch to autoloading so we don't load up all this code upfront.

SamSaffron:
This is looking fine to me ... thanks @clupprich

lib/prometheus_exporter/server.rb
clupprich: Are you thinking of zeitwerk? I can give that a try.

SamSaffron: was thinking of something far simpler like: 

https://github.com/ruby/ruby/blob/d92f09a5eea009fa28cd046e9d0eb698e3d94c5c/lib/bundler.rb#L42

vadasambar:
Any idea when this would be released?

SamSaffron:
No worries, will push a new version out today, thanks for the reminder.

Issue #172, opened 2021-06-17
Add hostname to puma instrumentation


codez:
Other than e.g. the process instrumentation, the puma metrics do not include a hostname. In our setup, multiple puma containers report towards the same prometheus exporter. Without the hostname, they overwrite each others values.

There is also no option to add custom labels to the puma instrumentation either. This would have allowed us to add a hostname label on our own. 

I am willing to work on a pull request for this. Which option do you prefer, always adding a hostname label, add custom labels or both?

SamSaffron:
I think it makes sense to add unconditionally

codez:
Oops, I'm sorry I don't quite understand your answer as a non-native English speaker. Do you think it's best to add both an option for custom labels and the hostname unconditionally?

By the way, I saw that while other instrumentations add the hostname, it is not propagated by the server but only used to distinguish different values. This should not be needed for the puma metrics IMHO.

SamSaffron:
Sorry, I was way too short there. 

I think it is fine to add hostname label for Puma unconditionally in a PR 

I am also not against adding support for custom label support in puma in a different PR.

codez:
That's ok, I will try to start work on it in the next couple of days.

bgmat:
Hi there,
Any plans to release this?
I really need it.
Thanks

PR #174, opened 2021-06-21
FEATURE: always scope puma metrics on hostname in collector (#172)

codez:
As discussed in #172, separate puma metrics by hostname and purge previous values.


SamSaffron:
This looks good, any chance you can add a test for this? If not the change looks pretty safe I can merge.

codez:
There you go.

SamSaffron (approved):

SamSaffron:
Looks good! Thanks.

PR #175, opened 2021-06-30
FIX: ThreadError Handling

keerthivaasan-kanagaraj:
Hi @SamSaffron

We observed ThreadError was raised from prometheus_exporter when sidekiq triggers SHUTDOWN EXCEPTION.It was due to some improper shutdowns.

As per ruby ThreadError will raise only when main thread is killed then we are trying to create child thread for that main thread

```
if (GET_VM()->main_thread->status == THREAD_KILLED)
rb_raise(rb_eThreadError, "can't alloc thread");
```

So need to catch exception in Prometheus gem to handle this error.

Ref. PR: (opentelemetry-ruby) https://github.com/open-telemetry/opentelemetry-ruby/pull/607

### DIFF
--- a/lib/prometheus_exporter/client.rb
+++ b/lib/prometheus_exporter/client.rb
@@ -184,6 +184,9 @@ module PrometheusExporter
           end
         end
       end
+    rescue ThreadError => e
+      raise unless e.message =~ /can't alloc thread/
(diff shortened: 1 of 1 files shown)


SamSaffron:
Seems good to me, a tiny bit fragile in case Ruby change the text on the exception but I guess that is the best we can do here.

Issue #83, opened 2019-08-15
use standard logging for errors instead of STDERR


kleinron:
If the target server is not accessible to the client, the following message repeats itself in STDERR:
```
Prometheus Exporter, failed to send message Connection refused - connect(2) for "localhost" port 9394
```
(the actual host and port might change, of course).

This outcome causes the standard error to be flooded with error messages.

Perhaps there's a way to inject either a stream object (with the same API as STDERR/STDOUT) or a logger object so that the hosting application could control the logging?

dorongutman:
Yes please, we are running a few environments without the server part (for example - a console) and in those cases there's no need for the prometheus exporter server, but it's impossible to work with these error messages.

I also think that there should be an option to disable the endless retry loop if the server isn't available. Maybe a configurable max retry attempts?

SamSaffron:
Sure. Happy to play with the delay here and to introduce a pluggable logger if anyone wants to try it out. This could be a warn and you could only be interested in errors.

PR #177, opened 2021-07-20
swap from hardcoded STDERR to logger

anujbiyani:
logger still defaults to STDERR, so there's no change unless a user passes in their own logger

~I wasn't sure how to unit test this, doesn't seem like logger code is usually unit tested but happy to add them if you'd like!~ [`75281d3` (#177)](https://github.com/discourse/prometheus_exporter/pull/177/commits/75281d399ff2c5cd25280315d957d431e5a62cad)

I think there are three affected code paths:
1. running via `bin/prometheus_exporter`
2. including the gem in another project and initializing a client
3. running a server like https://github.com/discourse/prometheus_exporter#single-process-mode

I tested 1+2 as thoroughly as I could think of. For 3 I'm not sure how to test it other than launching the server (which works fine).

fixes https://github.com/discourse/prometheus_exporter/issues/83

### DIFF
--- a/lib/prometheus_exporter/instrumentation/active_record.rb
+++ b/lib/prometheus_exporter/instrumentation/active_record.rb
@@ -7,11 +7,6 @@ module PrometheusExporter::Instrumentation
 
     def self.start(client: nil, frequency: 30, custom_labels: {}, config_labels: [])
 
-      config_labels.map!(&:to_sym)
-      validate_config_labels(config_labels)
-
-      active_record_collector = new(custom_labels, config_labels)
-
       client ||= PrometheusExporter::Client.default
 
       # Not all rails versions support connection pool stats
@@ -19,6 +14,11 @@ module PrometheusExporter::Instrumentation
         client.logger.error("ActiveRecord connection pool stats not supported in your rails version")
         return
       end
+
+      config_labels.map!(&:to_sym)
+      validate_config_labels(config_labels)
+
+      active_record_collector = new(custom_labels, config_labels)
 
       stop if @thread
 
--- a/lib/prometheus_exporter/instrumentation/active_record.rb
+++ b/lib/prometheus_exporter/instrumentation/active_record.rb
@@ -7,18 +7,18 @@ module PrometheusExporter::Instrumentation
 
     def self.start(client: nil, frequency: 30, custom_labels: {}, config_labels: [])
 
-      # Not all rails versions support coonection pool stats
-      unless ::ActiveRecord::Base.connection_pool.respond_to?(:stat)
-        STDERR.puts("ActiveRecord connection pool stats not supported in your rails version")
-        return
-      end
-
       config_labels.map!(&:to_sym)
       validate_config_labels(config_labels)
 
       active_record_collector = new(custom_labels, config_labels)
 
       client ||= PrometheusExporter::Client.default
+
+      # Not all rails versions support connection pool stats
+      unless ::ActiveRecord::Base.connection_pool.respond_to?(:stat)
+        client.logger.error("ActiveRecord connection pool stats not supported in your rails version")
+        return
+      end
 
       stop if @thread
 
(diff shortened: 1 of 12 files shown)


lib/prometheus_exporter/instrumentation/active_record.rb
```diff
@@ -7,19 +7,19 @@ class ActiveRecord
...
       active_record_collector = new(custom_labels, config_labels)
 
       client ||= PrometheusExporter::Client.default
 
+      # Not all rails versions support connection pool stats
+      unless ::ActiveRecord::Base.connection_pool.respond_to?(:stat)
+        client.logger.error("ActiveRecord connection pool stats not supported in your rails version")
+        return
+      end
+
```
SamSaffron, line 22: Are we sure the other stuff is supported here on all Rails versions, I worry that this fix may raise even more support issues when it is partially collecting?

SamSaffron:
Overall this looks great and complete. 

Yeah I think it makes sense to have some minimal testing here, you can just implement a test logger and confirm at least that it is passed in and usable?

lib/prometheus_exporter/instrumentation/active_record.rb
anujbiyani: I'm not sure I follow; the only change here is I moved this code from a few lines higher down to here that way we could use `client.logger`.

I looked at the code that is now executing (that wouldn't have prior to my change) and it looks like it's all internal setup for this gem and not dependent on Rails at all. Code that might now execute before the guard clause instead of after:
```ruby
      config_labels.map!(&:to_sym) # harmless
      validate_config_labels(config_labels) # internal validation. this could now raise whereas before it wouldn't even execute

      active_record_collector = new(custom_labels, config_labels) # assigns ivars, harmless

      client ||= PrometheusExporter::Client.default # instantiates client but doesn't actually do any work, harmless
```

The `raise` could be a problem if an app that doesn't support `stat` is _also_ passing in invalid config labels as now that app would `raise`. I'll move that code further down so it preserves the current behavior.

anujbiyani: This should better preserve the existing behavior: [`03a3714` (#177)](https://github.com/discourse/prometheus_exporter/pull/177/commits/03a37145fb1694b487bf7b79d4befc4189357f34)

anujbiyani:
Thanks for the push, I think I was just being lazy. This seems like a useful test to add: [`75281d3` (#177)](https://github.com/discourse/prometheus_exporter/pull/177/commits/75281d399ff2c5cd25280315d957d431e5a62cad)

test/client_test.rb
```diff
@@ -51,4 +51,17 @@ def test_standard_values
     summary_metric = client.register(:summary, 'summary_metric', 'helping', expected_quantiles)
     assert_equal(expected_quantiles, summary_metric.standard_values('value', 'key')[:opts])
   end
+
+  def test_overriding_logger
+    mock_logger = Minitest::Mock.new
```
SamSaffron, line 56: I know it is a lot to ask, but is there any way to do this test in a less mocky way? perhaps pass our logger in, then trigger a condition that causes some logging? 

Maybe we expose max_queue_size, so we dont need to reach in with instance_var_set?

anujbiyani: I can do that! I wasn't sure when I wrote this whether or not exposing `max_queue_size` would be kosher, but I can add it to the initializer and then remove `instance_var_set`.

anujbiyani:
Looks like Appraisal has a known issue that matches the current build failure, with a possible fix this Friday: https://github.com/thoughtbot/appraisal/pull/184#issuecomment-884087923

test/client_test.rb
anujbiyani: [`2d20be8` (#177)](https://github.com/discourse/prometheus_exporter/pull/177/commits/2d20be835dc00669b3d68b7decb7c24240a974ee)

SamSaffron:
This is a great PR and change @anujbiyani, thank you! 

If you feel like doing any triage on other open bugs we have I would be delighted!

anujbiyani:
@SamSaffron thanks for all your work on this gem!

theorician:
@SamSaffron Hi, would you mind releasing a point version with this change in? We've been waiting for this eagerly and would love to get this feature in to minimise some of the output spam. 

@anujbiyani Thanks for this change!

SamSaffron:
Sure ... will post a release today.

PR #181, opened 2021-08-17
Remove job labels from DelayedJob queue gauges

### LINKED ISSUE #180: DelayedJob gauges are queue counts but labelled with job names
The gauge metrics for DelayedJob (jobs pending, jobs enqueued) are counted for all jobs in a queue, but are labelled with the job name. This causes a gauge to be created for each job, but the gauge is only updated when that particular job runs. This means it isn't possible to query Prometheus to determine the actual number of jobs pending or enqueued, because you can't tell which gauge was most recently updated.

jferris:
The gauges count all jobs for a queue, not just jobs of that particular class.

Resolves #180.

### DIFF
--- a/lib/prometheus_exporter/server/delayed_job_collector.rb
+++ b/lib/prometheus_exporter/server/delayed_job_collector.rb
@@ -19,22 +19,21 @@ module PrometheusExporter::Server
     end
 
     def collect(obj)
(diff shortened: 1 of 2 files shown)


SamSaffron:
Looks good, thanks!

Issue #182, opened 2021-08-23
rename metrics to match prometheus official naming convention


Shimokuni:
Hello.

In this gem, there are some metrics which have `_total` suffix name but actually its metrics type is `Gauge`.

e.g. Unicorn process metrics

Type | Name | Description
-- | -- | --
Gauge | unicorn_workers_total | Number of unicorn workers
Gauge | unicorn_active_workers_total | Number of active unicorn workers
Gauge | unicorn_request_backlog_total | Number of requests waiting to be processed by a unicorn worker

https://github.com/discourse/prometheus_exporter#metrics-collected-by-unicorn-instrumentation

According to [prometheus official guide for metric names](https://prometheus.io/docs/practices/naming/#metric-names), the metrics which are named with `_total` suffix must be `Counter`, not `Gauge`.

This can be a problem when integrating with other services which infers the metric type based on prometheus naming conventions. 
e.g.  New Relic prometheus remote write
https://docs.newrelic.com/docs/integrations/prometheus-integrations/install-configure-remote-write/set-your-prometheus-remote-write-integration/#mapping

Of course we can use relabelConfig, but It takes effort...

What do you think about renaming the metrics to match the formal Promethean naming convention?
Can I write PR?
It will be breaking change, but I think it's the right change.

SamSaffron:
Sure I am fine with a rename here, I guess we need a major version change cause we could break people. 

`unicorn_workers`, `unicorn_active_workers` `unicorn_request_backlog`, be sure to add a note to changelog and review the readme for accuracy.

Shimokuni:
PR ready. https://github.com/discourse/prometheus_exporter/pull/184
Could you review?

PR #184, opened 2021-08-24
rename metrics to match prometheus official naming conventions

Shimokuni:
closes https://github.com/discourse/prometheus_exporter/issues/182

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+1.0.0 - dd-mm-yyyy
+
+- Breaking change: rename metrics to match prometheus official naming conventions (See https://prometheus.io/docs/practices/naming/#metric-names)
+
 0.8.1 - 04-08-2021
 
 - FEATURE: swap from hardcoded STDERR to logger pattern (see README for details)
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@ Prometheus Exporter allows you to aggregate custom metrics from multiple process
 To learn more see [Instrumenting Rails with Prometheus](https://samsaffron.com/archive/2018/02/02/instrumenting-rails-with-prometheus) (it has pretty pictures!)
 
 * [Requirements](#requirements)
+* [Migrating from v0.x](#migrating-from-v0.x)
 * [Installation](#installation)
 * [Usage](#usage)
   * [Single process mode](#single-process-mode)
@@ -36,6 +37,12 @@ To learn more see [Instrumenting Rails with Prometheus](https://samsaffron.com/a
 ## Requirements
 
 Minimum Ruby of version 2.5.0 is required, Ruby 2.4.0 is EOL as of 2020-04-05
+
+## Migrating from v0.x
+
+There are some major changes in v1.x from v0.x.
+
+- Some of metrics are renamed to match [prometheus official guide for metric names](https://prometheus.io/docs/practices/naming/#metric-names). (#184)
 
 ## Installation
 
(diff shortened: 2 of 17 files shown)


CHANGELOG
```diff
@@ -1,3 +1,7 @@
+1.0.0 - dd-mm-yyyy
```
Shimokuni, line 1: FIX ME

Shimokuni (commented):
May I edit [version.rb](lib/prometheus_exporter/version.rb)?

README.md
```diff
@@ -554,14 +561,14 @@ PrometheusExporter::Instrumentation::Resque.start
...
-| ---   | ---                    | ---                                    |
-| Gauge | `processed_jobs_total` | Total number of processed Resque jobs  |
-| Gauge | `failed_jobs_total`    | Total number of failed Resque jobs     |
-| Gauge | `pending_jobs_total`   | Total number of pending Resque jobs    |
-| Gauge | `queues_total`         | Total number of Resque queues          |
-| Gauge | `workers_total`        | Total number of Resque workers running |
-| Gauge | `working_total`        | Total number of Resque workers working |
+| Type  | Name             | Description                            |
+| ---   | ---              | ---                                    |
+| Gauge | `processed_jobs` | Total number of processed Resque jobs  |
```
SamSaffron, line 566: qw have to fix these name `resque_processed_jobs` etc.

SamSaffron:
thanks, looks like you are making good progress. We do need to also improve metric names and make them more consistent, can you have a look?

README.md
Shimokuni: Fixed. https://github.com/discourse/prometheus_exporter/pull/184/commits/ab91609fb46760f4ace046ce68753f379600465a

Shimokuni:
I think this is all fixed now. https://github.com/discourse/prometheus_exporter/pull/184/commits/a845e548b62deffb9f2a61a233f9dca8dcd9a192
And I added some missing tests for unicorn. https://github.com/discourse/prometheus_exporter/pull/184/commits/9669a3b7fb8b44d8cbff1aba2add00751dbd4d9a

---

> We do need to also improve metric names and make them more consistent, can you have a look?

For example, do you mean to fix the inconsistency that resque has a gauge type metric called `resque_failed_jobs`, while sidekiq has a counter type metric called `sidekiq_failed_jobs_total`?

SamSaffron:
This is a very nice change, going to hold off a bit on a release cause it has potential to break quite a few people, let's let it bake for 2-4 weeks.

salzig:
Would've been awesome if changes like those get a graceperiod where both names are provided. We ended up in a situation where our dashboards are working for one or the other environemnt (as in stage/production), cause our deployment cycles are sadly not as fast as we wish for (Customer does it's own QA additionally, and very slow).

SamSaffron:
@salzig kind of impossible to do this without some very bad side effects (exploding metric counts) 

The change was flagged as a major change we did major version bump here.

salzig:
@SamSaffron i don't see a problem with the name change in the major version. Maybe our setup is "strange", but name changes on one day to another are quite annoying, as we have to choose between showing the old or the new name, or query both metrics in our panels, as our Grafana installation is central across all environments. 

Sorry, overall my last message was written in frustration, cause another metric (http request duration) was also renamed, and additionally lost a label which we used (status). Keeping a old version, isn't really an option. So we need to figure out how to proceed. 

Thanks to all of you for providing and maintaining this as open source.

PR #185, opened 2021-08-30
Fix logger for web_server

vin01:
Hi, 

it looks like this has been broken since https://github.com/discourse/prometheus_exporter/commit/c3e26a827b59c225a2ffb34eec57d8541e6ee97f

I just noticed this when there was a silent `500` because of a [timeout](https://github.com/discourse/prometheus_exporter/blob/master/lib/prometheus_exporter/server/web_server.rb#L124) and got:

```
 ERROR NameError: undefined local variable or method `logger' for #<PrometheusExporter::Server::WebServer:0x000055f23ed4db50>\nDid you mean?  @logger
 ```

### DIFF
--- a/lib/prometheus_exporter/server/web_server.rb
+++ b/lib/prometheus_exporter/server/web_server.rb
@@ -88,7 +88,7 @@ module PrometheusExporter::Server
           @collector.process(block)
         rescue => e
           if @verbose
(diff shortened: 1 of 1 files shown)


SamSaffron:
nice catch!

PR #186, opened 2021-09-13
Allow collecting web metrics as histograms

### LINKED ISSUE #92: Unable to aggregate web metrics due to use of summary
More of a question about design decisions rather than a bug, but sort of a bug.

As it stands, the web collector defines summary metrics for durations.

As noted in the Prometheus documentation, summary metrics can't be correctly aggregated across multiple labels (the main use case here being across multiple servers).

I'm wondering why summary metrics are used here instead of histogram metrics - the latter would allow the correct aggregation of statistics between servers (allowing calculation of overall percentiles etc)?

jferris:
Web metrics are currently reported as summaries, which provides precise
data for each controller, action, and node. However, timings from
low-throughput actions will cause an average to be wildly inaccurate,
which means aggregating across controllers, actions, and nodes is not
currently possible.

This adds an option to the middleware which instructs the collector to
use histogram metrics rather than summaries, which allows estimating
aggregates using histogram_quantile.

Resolves #92.

### DIFF
--- a/lib/prometheus_exporter/middleware.rb
+++ b/lib/prometheus_exporter/middleware.rb
@@ -9,6 +9,7 @@ class PrometheusExporter::Middleware
   def initialize(app, config = { instrument: true, client: nil })
     @app = app
     @client = config[:client] || PrometheusExporter::Client.default
+    @mode = config.fetch(:mode, :summary)
 
     if config[:instrument]
       if defined? Redis::Client
@@ -41,7 +42,8 @@ class PrometheusExporter::Middleware
       type: "web",
       timings: info,
       queue_time: queue_time,
-      default_labels: default_labels(env, result)
+      default_labels: default_labels(env, result),
+      options: { mode: @mode }
     }
     labels = custom_labels(env)
     if labels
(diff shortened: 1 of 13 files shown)


lib/prometheus_exporter/middleware.rb
```diff
@@ -41,7 +42,8 @@ def call(env)
       type: "web",
       timings: info,
       queue_time: queue_time,
-      default_labels: default_labels(env, result)
+      default_labels: default_labels(env, result),
+      options: { mode: @mode }
```
SamSaffron, line 46: I wonder if we can avoid sending this piece of text with every metric? Maybe we use a different way of initializing that sets up the collector so it know it is getting historams vs summaries?

SamSaffron:
Hmm, tricky, I think a startup flag is best here, I like the idea of
dynamic config but it introduces a new security concern

jferris:
@SamSaffron - I pushed an update using a `--histogram` flag rather than an option for the middleware.

SamSaffron (approved):

SamSaffron:
This is great, thanks @jferris

svenjr:
When will this get released? I was trying to implement it using v0.8.1 of the gem but now see this was implemented after that release. Is there a release planned? Thanks

PR #188, opened 2021-09-21
Add back username and port as AR configuration values

### LINKED ISSUE #187: ALLOWED Config Labels is inaccurate for Rails 6.1
IN the code the list of allowed config labels for ActiveRecord are `username`, `database`, `host` and `port`. These are fetched from an `ActiveRecord::DatabaseConfigurations::HashConfig` object using `public_send`. Hence they expect a public method to return those values.

In Rails 6.1.4.1 (and I believe from Rails 6.1.0.rc1 onward) only 2 of those methods exist in Rails - `database` and `host`.

So in fact `username` and `port` - while allowed - return an `NoMethodError` in place of the valued content.

tjchambers:
In Rail 6.1 the information return is from an `ActiveRecord::DatabaseConfigurations::HashConfig` object, with methods for some but not all of the configuration attributes. This fixes a bug wheree `database` and `host` had getter methods, but the `username` and `port` values did not. All values are now just fetched from the instance variable configuration_hash directly.

This fixes #187

### DIFF
--- a/lib/prometheus_exporter/instrumentation/active_record.rb
+++ b/lib/prometheus_exporter/instrumentation/active_record.rb
@@ -87,7 +87,7 @@ module PrometheusExporter::Instrumentation
           .map { |k, v| [k.to_s.dup.prepend("dbconfig_"), v] }.to_h)
       else
         @metric_labels.merge(pool_name: pool.db_config.name).merge(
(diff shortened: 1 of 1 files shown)


SamSaffron:
I worry a bit about cross compatibility here. What is going to happen with earlier versions of rails?

tjchambers:
You separated the pre 6.1 code out, which is all that I changed.

SamSaffron:
Happy to try this, any idea why CI is failing?

PR #192, opened 2021-11-19
FEATURE: SidekiqProcess metrics

mrsimo:
Hi again! 👋🏻

We have this in our codebase and thought it might be interesting for everyone else. We have setup alerts for our Sidekiq deployments based on the latency metric coming from the `SidekiqQueue` instrumentation, but it was hard to see if we were overprovisioned or what, and we wanted to be able to have metrics on utilization (busy vs total).

The new metrics are based on Sidekiq's [`ProcessSet`](https://github.com/mperham/sidekiq/wiki/API#processes) class, and expose what it returns. It's a bunch of labels, but they seem useful and interesting enough.

I'd still need to add something to the README before the PR can be merged (and any other feedback, of course), but thought I'd open the PR early to see if this was interesting or not 😄


SamSaffron:
It sounds good to me, once you add the documentation! I can certainly see it as useful.

mrsimo:
Added a bit to the README. Not sure if more is necessary, advice is welcome 😄

mrsimo:
In case this is relevant at all, just tried this branch in our staging environment and everything seems to be in order.

Is anything else missing? Is an entry in the Changelog necessary? Feel free to add it if necessary, I think we're in opposite sides of the world 😆 

Thanks a lot Sam!

SamSaffron:
Looking good to me, we are due for a release, I can take the changelog!

Thanks @mrsimo !

PR #194, opened 2021-11-23
FEATURE: SidekiqStats metrics

mrsimo:
Hi again! Second PR today, sorry for the noise!

This is the last custom collector we have. I wasn't sure if it'd be as useful as the others, but we like having this information in Prometheus, so I thought it might be useful to others as well.

This collector just fetches [`Sidekiq::Stats`](https://github.com/mperham/sidekiq/wiki/API#stats) and exposes the numbers. No labels or anything involved.

Thank you!


lib/prometheus_exporter/server/sidekiq_stats_collector.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    SIDEKIQ_STATS_GAUGES = {
+      'dead_size' => 'Size of dead the queue',
+      'enqueued' => 'Number of enqueued jobs',
+      'failed' => 'Number of failed jobs',
+      'processed' => 'Total number of processed jobs',
+      'processes_size' => 'Number of processes',
+      'retry_size' => 'Size of the retries queue',
+      'scheduled_size' => 'Size of the scheduled queue',
+      'workers_size' => 'Number of jobs being worked',
```
SamSaffron, line 15: hmm this description seems a bit off. Perhaps:

"Number of jobs actively being processed"

README.md
```diff
@@ -404,6 +404,18 @@ Sidekiq.configure_server do |config|
 end
 ```
 
+To monitor general Sidekiq stats:
```
SamSaffron, line 407: A note about how these stats are "global" is probably in order, cause if you are collecting on multiple nodes that are working the same queue you will get dupes.

SamSaffron:
Overall seems fine to me, only a few small comments.

lib/prometheus_exporter/server/sidekiq_stats_collector.rb
mrsimo: 👍🏻 https://github.com/discourse/prometheus_exporter/pull/194/commits/dee7768fac9fdd31782f45ef9ba53aee6251fcd6

README.md
mrsimo: Not sure if this is overkill and you'd rather keep the previous format, but in https://github.com/discourse/prometheus_exporter/pull/194/commits/951c98a7918c3cec168126538cb90dd837b0ca50 I've rewritten this section a little bit, with the idea of giving an initial example with everything necessary, some explanations, and mentioning the option of running a long-lived process for `SidekiqQueue` and `SidekiqStats`. 

I think this would help people trying to use this gem. Thank you!

SamSaffron:
Yes, very much like your changes, thanks @mrsimo

PR #195, opened 2021-11-29
Trivial date format fixes

ujihisa:
## Problem

Since dd-mm-yyyy is ambiguous, it's hard to read, particularly when the dd is less than or equal to 12. It's also not commonly used.

## Solution

A standard such as [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) can be helpful, not only because it's commonly used, but it's also less confusing.

dd/mm/yyyy or mm/dd/yyyy are more common, but it still does not solve the ambiguous problem.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,4 @@
-1.0.0 - 23-11-2021
+1.0.0 - 2021-11-23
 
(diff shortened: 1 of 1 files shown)


SamSaffron:
sure, seems fine to me, thanks!

PR #197, opened 2021-12-03
fix dead:false ignored in sidekiq death handler

alexmorten:
You can prevent a sidekiq job form going into the dead-set by either setting `dead: false` or by setting `retry: false` ([wiki](https://github.com/mperham/sidekiq/wiki/Error-Handling#dead-set)). 

It seems like the `dead: false` option was overlooked in #76 😅


tgxworld:

PR #202, opened 2021-12-14
FEATURE: Add labels to preflight CORS requests

ramongtx:
Adds support for [rack-cors gem](https://github.com/cyu/rack-cors), so that Preflight requests don't get labeled as `controller="other"` and `action="other"` and instead get `controller="preflight"` and `action="/my/path"`.

### DIFF
--- a/lib/prometheus_exporter/middleware.rb
+++ b/lib/prometheus_exporter/middleware.rb
@@ -54,10 +54,18 @@ class PrometheusExporter::Middleware
   def default_labels(env, result)
     status = (result && result[0]) || -1
     params = env["action_dispatch.request.parameters"]
+    cors = env["rack.cors"]
     action = controller = nil
     if params
       action = params["action"]
       controller = params["controller"]
+    elsif cors && cors.respond_to?(:preflight?) && cors.preflight?
+      # if the Rack CORS Middleware identifies the request as a preflight request,
+      # the stack doesn't get to the point where controllers/actions are defined
+      action = "preflight"
+      controller = "preflight"
+    end
+
     end
 
     {


lib/prometheus_exporter/middleware.rb
```diff
@@ -58,6 +58,13 @@ def default_labels(env, result)
     if params
       action = params["action"]
       controller = params["controller"]
+    elsif env["rack.cors"] && env["rack.cors"].preflight?
```
SamSaffron, line 61: maybe we check that it responds to `preflight?` just in case?

ramongtx: Good idea. I've added a `respond_to?` for this and also abstained from using the `REQUEST_PATH` as action, since that could have a huge impact on the cardinality of the metrics.

lib/prometheus_exporter/middleware.rb
```diff
@@ -58,6 +58,13 @@ def default_labels(env, result)
     if params
       action = params["action"]
       controller = params["controller"]
+    elsif env["rack.cors"] && env["rack.cors"].respond_to?(:preflight?) && env["rack.cors"].preflight?
```
SamSaffron, line 61: last minor one, no need to look up in the hash 3 times. might as well store in a local var then do the tests on that.

Except for that seems fine.

ramongtx: Changed that too, thanks!

lib/prometheus_exporter/middleware.rb
```diff
@@ -54,10 +54,18 @@ def call(env)
   def default_labels(env, result)
     status = (result && result[0]) || -1
     params = env["action_dispatch.request.parameters"]
+    cors = env["rack.cors"]
     action = controller = nil
     if params
       action = params["action"]
       controller = params["controller"]
+    elsif cors && cors.respond_to?(:preflight?) && cors.preflight?
```
SamSaffron, line 62:
```suggestion
    elsif (cors = env["rack.cors"]) && cors.respond_to?(:preflight?) && cors.preflight?
```

Minor, but might as well push back the lookup.

```
irb(main):002:0> if (bar=foo[:bar]) && bar == true; puts "hi"; end
hi
```

SamSaffron (approved):

PR #207, opened 2022-01-17
feat(histogram): buckets

kayn1:
## What?

Allow to specify values for default buckets in histogram.

## Why?
Would be good to have ability to specify your own values for buckets. There is a similar ability for labels.

### DIFF
--- a/lib/prometheus_exporter/metric/histogram.rb
+++ b/lib/prometheus_exporter/metric/histogram.rb
@@ -5,6 +5,16 @@ module PrometheusExporter::Metric
 
     DEFAULT_BUCKETS = [0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5.0, 10.0].freeze
 
+    @default_buckets = nil if !defined?(@default_buckets)
+
+    def self.default_buckets
+      @default_buckets || DEFAULT_BUCKETS
+    end
+
+    def self.default_buckets=(buckets)
+      @default_buckets = buckets
+    end
+
     def initialize(name, help, opts = {})
       super(name, help)
       @buckets = (opts[:buckets] || DEFAULT_BUCKETS).sort.reverse
--- a/test/metric/histogram_test.rb
+++ b/test/metric/histogram_test.rb
@@ -139,5 +139,19 @@ module PrometheusExporter::Metric
 
       assert_equal(histogram.to_h, { name: "bob", family: "skywalker" } => { "count" => 3, "sum" => 1.79 })
     end
+
+    it 'supports default buckets' do
+      assert_equal(Histogram::DEFAULT_BUCKETS, [0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5.0, 10.0])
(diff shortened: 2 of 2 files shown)


lib/prometheus_exporter/metric/histogram.rb
```diff
@@ -5,6 +5,16 @@ class Histogram < Base
...
+      @default_buckets || DEFAULT_BUCKETS
+    end
+
+    def self.default_buckets=(buckets)
+      @default_buckets = buckets
+    end
+
     def initialize(name, help, opts = {})
       super(name, help)
       @buckets = (opts[:buckets] || DEFAULT_BUCKETS).sort.reverse
```
SamSaffron, line 20: how is this going to work if it is not used here? 

Is there precedence in the API for a default override like this?

kayn1: Good spot! I have just pushed changes in 31969581eadbec2c98c66177a4b72736f9aa3d16
Default buckets will be overwritten by this one specified in `opt[:buckets]` :+1:

kayn1: @SamSaffron Would you take a look? 🙏

test/metric/histogram_test.rb
```diff
@@ -139,5 +139,39 @@ module PrometheusExporter::Metric
...
+      custom_buckets = [0.005, 0.1, 1, 2, 5, 10]
+      Histogram.default_buckets = custom_buckets
+
+      assert_equal(Histogram.default_buckets, custom_buckets)
+
+      Histogram.default_buckets = Histogram::DEFAULT_BUCKETS
+    end
+
+    it 'uses the default buckets for instance' do
+      assert_equal(histogram.buckets.sort, Histogram::DEFAULT_BUCKETS)
```
SamSaffron, line 158: why is the sort here?

test/metric/histogram_test.rb
```diff
@@ -139,5 +139,39 @@ module PrometheusExporter::Metric
...
+    end
+
+    it 'uses the the custom default buckets for instance' do
+      custom_buckets = [0.005, 0.1, 1, 2, 5, 10]
+      Histogram.default_buckets = custom_buckets
+
+      assert_equal(histogram.buckets.sort, custom_buckets)
```
SamSaffron, line 165: and here?

SamSaffron:
I think this is fine, do we have anything to educate users about this pattern in the readme? can you add that?

test/metric/histogram_test.rb
kayn1: I wrote the test values in the increasing order so it was opposite of 
https://github.com/discourse/prometheus_exporter/blob/main/lib/prometheus_exporter/metric/histogram.rb#L10
I can change the order and not use `sort`

kayn1:
> I think this is fine, do we have anything to educate users about this pattern in the readme? can you add that?

@SamSaffron Sure, I will add description and examples to the readme. 👍

test/metric/histogram_test.rb
kayn1: Done 👍

kayn1:
@SamSaffron 🆙

SamSaffron (approved):

PR #208, opened 2022-01-21
Add per worker custom labels

TLischkaRemerge:
This adds the ability for sidekiq worker classes to have a self.custom_labels field which lets them set custom labels to be attached to metrics they export. This makes it easier to alerts based on the metrics for specific workers or groups of workers

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ on:
       - main
   pull_request:
   schedule:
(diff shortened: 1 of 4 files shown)


SamSaffron:
I am good to merge this, but there is a merge conflict and also looks like we need a README update

SamSaffron (approved):

SamSaffron:
thanks heaps @TLischkaRemerge !

PR #211, opened 2022-01-27
Rename metric http_duration_seconds

### LINKED ISSUE #210: Rename http_duration_seconds metric to http_request_duration_seconds
Hello,
Same thing as in #182 , we noticed that [prometheus](https://prometheus.io/docs/practices/naming/#metric-names) suggest to have `http_request_duration_seconds` instead of just `http_duration_seconds`.

Would you be open to this renaming, and bumping a new release with this breaking change?

Thank you!

MaximeD:
Official prometheus documentation refers to this metric as
`http_request_duration_seconds`.

See: https://prometheus.io/docs/practices/naming/#metric-names

Closes #210

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+Unreleased
+
+- BREAKING: rename all `http_duration` metrics to `http_request_duration` to match prometheus official naming conventions (See https://prometheus.io/docs/practices/naming/#metric-names).
+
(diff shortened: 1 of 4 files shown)


SamSaffron:
It is a scary change, but yeah I do support it, the naming is better.

SamSaffron (approved):

PR #148, opened 2020-12-19
Better extendability

graudeejs:
Hello!

I'd like to use prometheus_exporter, however I didn't like the way it was adding default labels.
My issue is that on my custom rack app controller and action always resolves to `other`.

I've modified prometheus_exporter in order to allow more flexibility when it comes to extendability.
This change allows me to create custom middleware by extending PrometheusExporter::Middleware and implement proper default tags (in my case, status, method and path. Similar how it's done in [prometheus-client](https://github.com/prometheus/client_ruby/blob/master/lib/prometheus/middleware/collector.rb#L69))

Please review my proposed change and merge, if deemed acceptable or please do let me know, what to improve.

Here's example metrics from my branch
```
# HELP ruby_http_requests_total Total HTTP requests from web app.
# TYPE ruby_http_requests_total counter
ruby_http_requests_total{path="/api/v1/persons",method="GET",status="200"} 1
ruby_http_requests_total{path="/api/v1/teams",method="GET",status="200"} 1
ruby_http_requests_total{path="/api/v1/persons/:id",method="PUT",status="200"} 1

# HELP ruby_http_duration_seconds Time spent in HTTP reqs in seconds.
# TYPE ruby_http_duration_seconds summary
ruby_http_duration_seconds{path="/api/v1/persons",method="GET",status="200",quantile="0.99"} 0.2886974339999142
ruby_http_duration_seconds{path="/api/v1/persons",method="GET",status="200",quantile="0.9"} 0.2886974339999142
ruby_http_duration_seconds{path="/api/v1/persons",method="GET",status="200",quantile="0.5"} 0.2886974339999142
ruby_http_duration_seconds{path="/api/v1/persons",method="GET",status="200",quantile="0.1"} 0.2886974339999142
ruby_http_duration_seconds{path="/api/v1/persons",method="GET",status="200",quantile="0.01"} 0.2886974339999142
ruby_http_duration_seconds_sum{path="/api/v1/persons",method="GET",status="200"} 0.2886974339999142
ruby_http_duration_seconds_count{path="/api/v1/persons",method="GET",status="200"} 1
ruby_http_duration_seconds{path="/api/v1/teams",method="GET",status="200",quantile="0.99"} 0.009880661998977303
ruby_http_duration_seconds{path="/api/v1/teams",method="GET",status="200",quantile="0.9"} 0.009880661998977303
ruby_http_duration_seconds{path="/api/v1/teams",method="GET",status="200",quantile="0.5"} 0.009880661998977303
[...]

PR #216, opened 2022-02-09
Do not include status label everywhere

MaximeD:
The `status` label is included in every metrics, and not only in
`http_requests_total`, even though the README states that
```
All metrics have a `controller` and an `action` label.
`http_requests_total` additionally has a (HTTP response) `status` label.
```

This can be a problem as this `status` label can quickly increase the
cardinality of the metrics.

---

This change of behavior was introduced in #148. But I believe this is not the best pattern as it is not useful to have it everywhere, and it can quickly blow up the cardinality. See those [slides about Containing your Cardinality](https://promcon.io/2019-munich/slides/containing-your-cardinality.pdf), especially page 14.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 Unreleased
 
+- FIX: all metrics are exposing status label, and not only `http_requests_total`
 - BREAKING: rename all `http_duration` metrics to `http_request_duration` to match prometheus official naming conventions (See https://prometheus.io/docs/practices/naming/#metric-names).
 
 1.0.1 - 2021-12-22
--- a/lib/prometheus_exporter/middleware.rb
+++ b/lib/prometheus_exporter/middleware.rb
@@ -36,11 +36,12 @@ class PrometheusExporter::Middleware
 
     result
   ensure
-
+    status = (result && result[0]) || -1
     obj = {
       type: "web",
       timings: info,
       queue_time: queue_time,
+      status: status,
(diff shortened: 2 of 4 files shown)


SamSaffron:
Is there any way to amend our internals so we only explicitly add the label where we need it vs removing it later on?

MaximeD:
Sure thing!
I reverted part of what was done in [#148](https://github.com/discourse/prometheus_exporter/pull/148/files#diff-fd5660ac2f44457dff6206a9508bac27d7e4368ddc7116e69c8311cfef952209L53), that is the `status` label is now again at the root of `obj` instead of being nested in `default_labels`.

SamSaffron:
cool, looks great!

SamSaffron (approved):

MaximeD:
Awesome!
By the way, do you plan to release a new version of this gem in the coming days?

Thank you again!

SamSaffron:
I have been a bit afraid to do a new release cause we had a few breaking changes ... that said .. got to rip off the cord at some point!

MaximeD:
Yes, only one way to battle test the new changes I guess!

salzig:
Another changes that makes harder to upgrade our dashboards. Thanks. 

We used to have a dashboard that shows duration per status code, as sometimes "sideeffects" (3rd Parties) can influence our app which is often seen very well isolated for a status code we're returning.

PR #219, opened 2022-02-23
FIX: ensure threads do not leak when #start is called

SamSaffron:
Periodic instrumentation now follows a general pattern where they
inherit from PeriodicStats.

This implementation ensures proper management of backing thread

1. All classes inheriting now have a .started? method to determine
if instrument is started

2. All classes implement a clean .stop method for stopping active
periodic instrumentation

3. Consumers of the class get to write less code repeat code due
to worker_loop

### DIFF
--- a/lib/prometheus_exporter/instrumentation/periodic_stats.rb
+++ b/lib/prometheus_exporter/instrumentation/periodic_stats.rb
@@ -1,0 +1,67 @@
+# frozen_string_literal: true
+
+module PrometheusExporter::Instrumentation
+  class PeriodicStats
+
+    def self.start(*args, **kwargs)
+      frequency = kwargs[:frequency]
+      client = kwargs[:client] || PrometheusExporter::Client.default
+
+      if !frequency
+        raise ArgumentError.new("Expected a value for frequency argument")
+      end
+
+      if !(Numeric === frequency)
+        raise ArgumentError.new("Expected frequency to be a number")
+      end
+
+      if frequency < 0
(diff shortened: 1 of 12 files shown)


martin-brennan (approved):
Nice refactor :) I see what you mean about the instrumenter tests, they look like they need lots of work. Minitest is so alien!

lib/prometheus_exporter/instrumentation/periodic_stats.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+module PrometheusExporter::Instrumentation
+  class PeriodicStats
+
+    def self.start(*args, **kwargs)
+      frequency = kwargs[:frequency]
+      client = kwargs[:client] || PrometheusExporter::Client.default
```
CvX, line 8: I think this would also work (not that it's a huge improvement 😉)

```suggestion
    def self.start(*args, frequency:, client: nil, **kwargs)
      client ||= PrometheusExporter::Client.default
```

CvX (approved):

lib/prometheus_exporter/instrumentation/periodic_stats.rb
SamSaffron: I think this works fine, we also would get to get rid of 1 assertion I think.

spiderpug:
We're encountering this issue https://github.com/discourse/prometheus_exporter/issues/222 after upgrading to 2.0.1.

Our mitigation: rollback to 2.0.0

PR #220, opened 2022-02-23
Allow gem to be used with ruby 2.5.5

TLischkaRemerge:
Since it is not possible to update ruby on the api servers at the
current time, prometheus_exporter needs to continue working with
the installed version (2.5.5) This change means that we can use
the recently added features without upgrading ruby

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ['2.6', '2.7', '3.0', '3.1']
+        ruby: ['2.5', '2.6', '2.7', '3.0', '3.1']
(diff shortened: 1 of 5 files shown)


SamSaffron:
Sorry I can not accept this, Ruby 2.5 was end of life 10 months ago. We have a strong policy of not supporting end of life Ruby which may have critical, unpatched, security vulnerabilities

TLischkaRemerge:
Oops, apologies for the noise here. I thought I'd made this one against our own fork as a messy and temporary solution but I mistakenly chose the upstream repo

SamSaffron:
No worries at all @TLischkaRemerge! good luck updating Ruby! 2.7 is better in almost all metrics than 2.5. 3.1 is also a very good release.

PR #225, opened 2022-02-25
FIX: unicorn instrumentation not required correctly

SamSaffron:
Runner.rb would crash due to only partially requiring unicorn
integration

Changes it so it is defer required

### DIFF
--- a/lib/prometheus_exporter/server/runner.rb
+++ b/lib/prometheus_exporter/server/runner.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require_relative '../client'
-require_relative '../instrumentation/unicorn'
 
 module PrometheusExporter::Server
   class RunnerException < StandardError; end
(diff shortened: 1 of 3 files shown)


lib/prometheus_exporter/server/runner.rb
```diff
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require_relative '../client'
-require_relative '../instrumentation/unicorn'
```
nattsw, line 4: Just to make sure.. no one else in this file needs anything from `'../instrumentation/unicorn'`, right?

discoursereviewbot: [SamSaffron posted](https://review.discourse.org/t/fix-unicorn-instrumentation-not-required-correctly-pr-225/32290/3):

I moved it to the spot where it used, only people using the cli with unicorn switches need it

spiderpug:
Thank you for the fast turnaround! 🤟

PR #226, opened 2022-02-25
Add /ping endpoint which only returns PONG

n-rodriguez:
So users can set an healthcheck in their `docker-compose.yml` file :

```yml
    healthcheck:
      test: ["CMD", "curl", "--silent", "--show-error", "--fail", "--max-time", "3", "http://0.0.0.0:9394/ping"]
      timeout: 3s
      interval: 10s
      retries: 5
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -952,6 +952,26 @@ You can also pass a log level (default is [`Logger::WARN`](https://ruby-doc.org/
 PrometheusExporter::Client.new(log_level: Logger::DEBUG)
 ```
 
+## Docker/Kubernetes Healthcheck
+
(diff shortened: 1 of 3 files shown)


SamSaffron:
:+1: 

I am fine with this but I would like a test and for it to be documented in the readme

n-rodriguez:
> I am fine with this but I would like a test and for it to be documented in the readme

sure!

n-rodriguez:
> I am fine with this but I would like a test and for it to be documented in the readme

done!

SamSaffron (approved):

PR #227, opened 2022-02-25
Pass msg hash to Sidekiq jobs custom_labels function

TLischkaRemerge:
[A previous PR](https://github.com/discourse/prometheus_exporter/pull/208) added the option to define a class method on Sidekiq worker classes. One thing that was accidentally left out is the ability for this function to use the arguments to an individual job run to generate the custom labels. I changed the code to pass the jobs hash to the custom_labels class function if it is defined.

The hash only gets passed if the custom_labels function expects > 0 arguments. This avoids a breaking change with the previous version where a custom_labels function on a worker class didn't take any arguments.

### DIFF
--- a/lib/prometheus_exporter/instrumentation/sidekiq.rb
+++ b/lib/prometheus_exporter/instrumentation/sidekiq.rb
@@ -16,7 +16,7 @@ module PrometheusExporter::Instrumentation
         job_is_fire_and_forget = job["retry"] == false
 
         worker_class = Object.const_get(job["class"])
-        worker_custom_labels = self.get_worker_custom_labels(worker_class)
+        worker_custom_labels = self.get_worker_custom_labels(worker_class, job)
 
         unless job_is_fire_and_forget
           PrometheusExporter::Client.default.send_json(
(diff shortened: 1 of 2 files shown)


SamSaffron:
the arity check does make me a bit nervous, can we put a TODO in the code base to remove it at some point?

TLischkaRemerge:
I know what you mean, I also felt quite apprehensive about adding it but I wasn't sure how to avoid a breaking change otherwise. Another option could be to leave it until the next major version is being prepared. Then it could just be a breaking change rather than having the workaround in place. What do you think?

SamSaffron:
I am ok just to keep the awkward  for now as long as you add a `#TODO remove when version X is released`

TLischkaRemerge:
I put a TODO comment to remove the arity check in the next major version: 3.0.0 Sorry it took a long time to get back to you on this one, caught covid in the meantime 😷

SamSaffron (approved):

SamSaffron:
Thanks @TLischkaRemerge

Issue #159, opened 2021-02-04
Redis instrumentation can cause infinite stack errors when combined with other instrumentation.


treyhyde:
When combined with the opentelemetry ruby gem, this instrumentation will cause an infinite stack and failure.

According to some, instrumenting with prepend would help various instrumentation play nicer together.
Alternatively or additionally, ability to disable various instrumentation (Redis, Pg, etc) would be of benefit.  In this exact case, this exporter is the more important of the two modules... but of the redis, db, etc instrumentation, the opentelemetry version would be my preference.

See
https://github.com/open-telemetry/opentelemetry-ruby/issues/579
https://github.com/open-telemetry/opentelemetry-ruby/issues/352
https://github.com/elastic/apm-agent-ruby/issues/379#issuecomment-717341248

etc

SamSaffron:
It is tricky, we went through a similar drill with rack-mini-profiler. prepend helps if everyone uses prepend but it can break stuff as well in other edge cases. 

My call here would be just to add switches to disable/enable bits of instrumentation you want and opt for the style of patching you prefer.

Happy to accept a PR here if you want to work on it.


https://github.com/MiniProfiler/rack-mini-profiler/pull/444

https://github.com/MiniProfiler/rack-mini-profiler/issues/437

Note instrumentation is already optional in that you have to `require 'prometheus_exporter/instrumentation'` but we could have a second `require 'prometheus_exporter/instrumentation_prepend'` or something like that for cross compatibility.

treyhyde:
Yeah, an option would be great.  The instrumentation is super valuable but I would prefer some finer grained controls rather than all on or all off.  Of course, I wouldn't care as much if it worked for my use case rather than pursue the micro-optimization and loss of metrics.   Really, the prepend option (theoretically) solves all my problems.   

Thanks team, we love this gem.

ball-hayden:
`prometheus_exporter` conflicting with `elastic-apm` is causing us grief - we're likely going to have to drop one or the other unless there's a sensible workaround (which would be a shame, because they both bring different value).

I guess a workaround for the timebeing is to monkeypatch https://github.com/discourse/prometheus_exporter/blob/e9511e40f47a8aaa4e94469955af802590133971/lib/prometheus_exporter/middleware.rb to not bind to Redis?

SamSaffron:
no need to monkey patch, you can just disable automatic instrumentation and write these lines yourself: 

[...]

PR #228, opened 2022-03-01
Ability to choose style of instrumentation

bcharna:
`prometheus_exporter` cannot reliably be used with other instrumentation libraries that use `prepend` to instrument methods as described in https://github.com/discourse/prometheus_exporter/issues/159. The goal of this PR is to provide the ability to have `prometheus_exporter` instrument using `prepend`, while still maintaining the default of `alias_method`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,14 @@ Ensure you run the exporter in a monitored background process:
 $ bundle exec prometheus_exporter
 ```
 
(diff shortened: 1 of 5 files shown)


SamSaffron:
what a wonderful PR, well documented, well tested, thanks @bcharna !

PR #230, opened 2022-03-09
FIX for ruby 3 incompatibility in sidekiq middleware, fixes #229

### LINKED ISSUE #229: Sidekiq middleware not Ruby >=3.0 compatible
The Sidekiq middleware is initialized with a keyword argument (https://github.com/discourse/prometheus_exporter/blob/main/lib/prometheus_exporter/instrumentation/sidekiq.rb#L36)
```ruby
def initialize(client: nil)
  @client = client || PrometheusExporter::Client.default
end
```
which goes against Sidekiq's base implementation of a middleware which suggests an options hash being passed in: https://github.com/mperham/sidekiq/wiki/Middleware#server-middleware
```ruby
class MyMiddleware::Server::ErrorLogger
  def initialize(options=nil)
    # options == { :foo => 1, :bar => 2 }
  end
  def call(worker, msg, queue)
    begin
      yield
    rescue => ex
      puts ex.message
    end
  end
end
```

While this is not a problem before Ruby 3.0, using the prometheus_exporter Sidekiq middleware with Ruby 3.0 or 3.1 will result in an `ArgumentError` because Sidekiq's middleware implementation only passes on `*args` (https://github.com/mperham/sidekiq/blob/main/lib/sidekiq/middleware/chain.rb#L158) and with the changes to positional/keyword arguments in Ruby 3.0 the method signature doesn't match any more.

I'm currently trying to put together a test case and PR to fix this in a backwards compatible way.

benzimmer:
See #229 for more details.

The test case might seem a bit weird, but I didn't find a more elegant way to reproduce Sidekiq's middleware initialization. So definitely open for suggestions.

- Sidekiq middleware chain for reference: https://github.com/mperham/sidekiq/blob/main/lib/sidekiq/middleware/chain.rb

### DIFF
--- a/lib/prometheus_exporter/instrumentation/sidekiq.rb
+++ b/lib/prometheus_exporter/instrumentation/sidekiq.rb
@@ -33,8 +33,8 @@ module PrometheusExporter::Instrumentation
       worker_class.respond_to?(:custom_labels) ? worker_class.custom_labels : {}
     end
 
-    def initialize(client: nil)
-      @client = client || PrometheusExporter::Client.default
+    def initialize(options = { client: nil })
+      @client = options.fetch(:client, nil) || PrometheusExporter::Client.default
     end
 
     def call(worker, msg, queue)
--- a/test/sidekiq_middleware_test.rb
+++ b/test/sidekiq_middleware_test.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+require 'minitest/stub_const'
+require_relative 'test_helper'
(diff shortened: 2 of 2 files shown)


SamSaffron:
@mperham what do you think of the workaround? I guess param handling changes between 2.7 / 2.8 / 3.0 are a bit of a sore point?

mperham:
Lgtm!

SamSaffron (approved):

SamSaffron:
thanks @benzimmer  !

PR #231, opened 2022-04-14
Parse delayed name in death_handler

bbxiao1:
This allows users who are still using the `delayed` extension to better monitor the dead queue by exporting `DelayedAction#foo` and `DelayedModel#action` instead of just `Sidekiq::Extensions::DelayedClass`.


SamSaffron (approved):

SamSaffron:
Thanks seems safe.

PR #232, opened 2022-05-17
Fix merge in web collector

renatolond:
Hi! Thanks for your work on prometheus_exporter :)
When upgrading prometheus_exporter, I noticed that status is no longer included in all requests, as per #216 . However, we would still like to include it in our custom labels, since we have some dashboards which use it.

When we tested including it ourselves, I got some errors from prometheus because there was an invalid metric. The metric had "status" twice. It seems that the issue is that the metric coming from the json serializer has `status` as a string, while the merge had a symbol.

This PR adds a test to check that the merge is correct and corrects the merge to use a string as well.


SamSaffron:
Oh ... nice catch. Thanks!

SamSaffron (approved):

PR #233, opened 2022-05-20
Order histogram buckets in increasing numerical order

clupprich:
The [docs on the Prometheus exposition format](https://prometheus.io/docs/instrumenting/exposition_formats/#histograms-and-summaries) say

> The buckets of a histogram and the quantiles of a summary must appear in increasing numerical order of their label values (for the le or the quantile label, respectively).

This hasn't been a problem with the Prometheus versions we've used up to now. However, I have a client that's migrating to GCP right now where they're using the [Managed Prometheus](https://cloud.google.com/stackdriver/docs/managed-prometheus) offering, which doesn't process the wrongly ordered histogram buckets.

Not sure what (custom?) Prometheus version GCP is using here (it's all pretty much hidden), though.

### DIFF
--- a/lib/prometheus_exporter/metric/histogram.rb
+++ b/lib/prometheus_exporter/metric/histogram.rb
@@ -19,7 +19,7 @@ module PrometheusExporter::Metric
 
     def initialize(name, help, opts = {})
       super(name, help)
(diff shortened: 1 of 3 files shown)


SamSaffron:
Nice change!

SamSaffron (approved):

PR #235, opened 2022-06-01
Update example custom collector to inherit from BaseCollector

geoffharcourt:
The documentation advises that a custom collector should inherit from
`PrometheusExporter::Server::BaseCollector` and the example in the docs
does so. The code example inherits from `Collector` instead. We found
this seemed to work in development but would behave unreliably in
production. This change updates the example to match the code snippet in
the Readme.


SamSaffron (approved):

SamSaffron:
Looks good, thanks for that!

PR #241, opened 2022-06-23
Revert breaking http rename

ayellapragada:
This change was made to follow prometheus conventions, but our albaik
manifests do not use this in the same way.

Because of this, we're temporarily reverting this one commit, or else
autoscaling could have issues when one deploy happens without either the
web or manifest changes.

Once autoscaling isn't needed, such as after hours, this fork can be
removed on any app that is using this and switch to the official
discourse gem.


SamSaffron:
no worries, sorry about the disruption here, I tried to signal pretty loudly by increasing version numbers. I agree with the change but also agree it is painful.

PR #250, opened 2022-10-27
FIX: correct instrumentation for Redis 5

SamSaffron:
Redis 5 gem introduces redis_client as the low level transport method

We no longer patch Redis gem if we detect it is version 5 and above and
instead use the supported middleware interface to patch it.

### DIFF
--- a/lib/prometheus_exporter/middleware.rb
+++ b/lib/prometheus_exporter/middleware.rb
@@ -6,12 +6,38 @@ require 'prometheus_exporter/client'
 class PrometheusExporter::Middleware
   MethodProfiler = PrometheusExporter::Instrumentation::MethodProfiler
 
+  module RedisInstrumenter
+    def call(_command, _config)
+      unless prof = Thread.current[:_method_profiler]
+        super
+      end
+
+      start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+      super
+    ensure
+      data = (prof[:redis] ||= { duration: 0.0, calls: 0 })
+      data[:duration] += Process.clock_gettime(Process::CLOCK_MONOTONIC) - start
+      data[:calls] += 1
+    end
+
+    alias_method :call_pipelined, :call
+  end
+
+  def apply_redis_client_middleware!
+    RedisClient.register(RedisInstrumenter)
+  end
+
   def initialize(app, config = { instrument: :alias_method, client: nil })
     @app = app
     @client = config[:client] || PrometheusExporter::Client.default
 
     if config[:instrument]
-      if defined? Redis::Client
+      if defined?(RedisClient)
+        apply_redis_client_middleware!
+      end
+      if defined?(Redis::VERSION) && (Gem::Version.new(Redis::VERSION) >= Gem::Version.new('5.0.0'))
+        # redis 5 support handled via RedisClient
+      elsif defined? Redis::Client
         MethodProfiler.patch(Redis::Client, [
           :call, :call_pipeline
         ], :redis, instrument: config[:instrument])
--- a/lib/prometheus_exporter/middleware.rb
+++ b/lib/prometheus_exporter/middleware.rb
@@ -6,27 +6,6 @@ require 'prometheus_exporter/client'
 class PrometheusExporter::Middleware
   MethodProfiler = PrometheusExporter::Instrumentation::MethodProfiler
 
-  module RedisInstrumenter
-    def call(_command, _config)
-      unless prof = Thread.current[:_method_profiler]
-        super
-      end
-
-      start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
-      super
-    ensure
-      data = (prof[:redis] ||= { duration: 0.0, calls: 0 })
-      data[:duration] += Process.clock_gettime(Process::CLOCK_MONOTONIC) - start
-      data[:calls] += 1
-    end
-
-    alias_method :call_pipelined, :call
-  end
-
-  def apply_redis_client_middleware!
-    RedisClient.register(RedisInstrumenter)
-  end
-
   def initialize(app, config = { instrument: :alias_method, client: nil })
     @app = app
     @client = config[:client] || PrometheusExporter::Client.default
@@ -140,6 +119,16 @@ class PrometheusExporter::Middleware
     value = env['HTTP_X_AMZN_TRACE_ID']
     value&.split('Root=')&.last&.split('-')&.fetch(1)&.to_i(16)
 
+  end
+
+  private
+
+  module RedisInstrumenter
+    MethodProfiler.define_methods_on_module(self, ["call", "call_pipelined"], "redis")
+  end
+
+  def apply_redis_client_middleware!
+    RedisClient.register(RedisInstrumenter)
   end
 
 end
--- a/test/middleware_test.rb
+++ b/test/middleware_test.rb
@@ -102,7 +102,6 @@ class PrometheusExporterMiddlewareTest < Minitest::Test
     redis.call("PING") # => "PONG"
     redis.call("PING") # => "PONG"
     results = PrometheusExporter::Instrumentation::MethodProfiler.stop
-    # redis client injects a HELLO preamble on reconnection, so we will see more than 2
     assert(2, results[:redis][:calls])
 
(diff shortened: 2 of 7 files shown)


lib/prometheus_exporter/middleware.rb
```diff
@@ -6,12 +6,38 @@
 class PrometheusExporter::Middleware
   MethodProfiler = PrometheusExporter::Instrumentation::MethodProfiler
 
+  module RedisInstrumenter
+    def call(_command, _config)
+      unless prof = Thread.current[:_method_profiler]
+        super
```
byroot, line 12: Don't you end up calling `super` twice here? That would emit the redis command twice.

lib/prometheus_exporter/middleware.rb
```diff
@@ -6,12 +6,38 @@
...
+
+      start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+      super
+    ensure
+      data = (prof[:redis] ||= { duration: 0.0, calls: 0 })
+      data[:duration] += Process.clock_gettime(Process::CLOCK_MONOTONIC) - start
+      data[:calls] += 1
+    end
+
+    alias_method :call_pipelined, :call
```
byroot, line 23: This will cause issues. If you have other middleware registered, you'll cause `call_pipelined` events to become `call` events.

I know it's a bit unfortunate, I only realized this after the API was published.

lib/prometheus_exporter/middleware.rb
```diff
@@ -6,12 +6,38 @@
...
+      unless prof = Thread.current[:_method_profiler]
+        super
+      end
+
+      start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+      super
+    ensure
+      data = (prof[:redis] ||= { duration: 0.0, calls: 0 })
+      data[:duration] += Process.clock_gettime(Process::CLOCK_MONOTONIC) - start
+      data[:calls] += 1
```
tgxworld, line 20: Will we be able to improve the API for `MethodProfiler` such that we don't have to re implement parts of what `MethodProfiler` does?

lib/prometheus_exporter/middleware.rb
```diff
@@ -6,12 +6,38 @@
...
+    end
+
+    alias_method :call_pipelined, :call
+  end
+
+  def apply_redis_client_middleware!
```
tgxworld, line 26: Minor nit but this can be `private`

test/middleware_test.rb
```diff
@@ -90,6 +90,19 @@ def test_amzn_trace_id_in_wrong_format
...
   end
 
+  def test_redis_5_patching
+    configure_middleware
+
+    redis_config = RedisClient.config(host: "127.0.0.1", port: 10)
+    redis = redis_config.new_pool(timeout: 0.5, size: 1)
+    PrometheusExporter::Instrumentation::MethodProfiler.start
+    redis.call("PING") # => "PONG"
+    redis.call("PING") # => "PONG"
```
tgxworld, line 100: I think `call_pipelined` has to be tested too.

lib/prometheus_exporter/middleware.rb
SamSaffron: oh yikes... will see if I can add a test for this, thanks for catching it

lib/prometheus_exporter/middleware.rb
SamSaffron: tricky cause of performance sensitivity, we will eat an extra method call for instrumentation :cry: not sure I love that

test/middleware_test.rb
SamSaffron: yeah the test is more complicated cause I need an additional middleware, will see what I can do.

lib/prometheus_exporter/middleware.rb
SamSaffron: oh implementation bug, will fix and add a test

lib/prometheus_exporter/middleware.rb
tgxworld: Technically, switching to the middleware for instrumentation is already adding an extra method call. I do wonder if an extra method call has any significance on performance in the big scheme of things. Even the profiling code itself is already making multiple additional method calls and I'm not convinced an extra method call will move the needle much.

SamSaffron:
@tgxworld / @byroot I think I have all the feedback addressed, let me know what you think

SamSaffron:
Oh tricky one @byroot ... what happens if a pipeline takes say N seconds to commit and get results back from Redis, how is the instrumenter going to pick it up?

test/middleware_test.rb
```diff
@@ -90,6 +90,42 @@ def test_amzn_trace_id_in_wrong_format
...
+    results = PrometheusExporter::Instrumentation::MethodProfiler.stop
+    # redis client injects a HELLO preamble on reconnection, so we will see more than 2
+    assert(results[:redis][:calls] >= 2, "expecting at least 2 calls")
```
tgxworld, line 104: I think this assertion is dangerous because it doesn't help us to catch regressions where we start to over increment method calls. For example, we may accidentally be increment twice instead of once and this test will still pass. 

On another note, I tried to add `redis.call("PING")` before `PrometheusExporter::Instrumentation::MethodProfiler.start` and the results of the call was still tracked in results even though method profiling should not have been started at the point in time. I think there is some state being leaked here.

SamSaffron: I think @byroot may need to provide an interface to allow us to tell redis-client ... hey ... no preamble please. Otherwise there are hello calls and no idea at all how many of them the implementation will make.

Will see if I can teach method profiler to behave a bit better, I have an idea for an improved test

SamSaffron: I am not sure there is anything I can do here, I tried a few things, unless we figure out a way of cleaning junking the HELLO calls.

Maybe I can register a HELLO remover middleware, so it does not noisy up the testing... will wait on byroot for guidance

byroot: If you use RESP2 protocol, there's no `HELLO` and that's what the `redis-rb` gem uses anyway: `RedisClient.new(protocol: 2)`

byroot:
> what happens if a pipeline takes say N seconds to commit and get results back from Redis, how is the instrumenter going to pick it up?

I'm not sure I understand what is tricky here.

```ruby
module MyInstru
  def call_pipelined(commands, config)
    start = Time.now
    result = super
    duration = Time.now - start # N seconds
    results
  end
end
```

SamSaffron:
Awesome thanks about the tip !

SamSaffron:
@tgxworld we are all done now, your comment is addressed

test/middleware_test.rb
```diff
@@ -90,6 +90,47 @@ def test_amzn_trace_id_in_wrong_format
...
+
+    # protocol 2 is the old redis protocol, it uses no preamble so you don't leak HELLO
+    # calls
+    redis_config = RedisClient.config(host: "127.0.0.1", port: 10, protocol: 2)
+    redis = redis_config.new_pool(timeout: 0.5, size: 1)
+    PrometheusExporter::Instrumentation::MethodProfiler.start
+    redis.call("PING") # => "PONG"
+    redis.call("PING") # => "PONG"
+    results = PrometheusExporter::Instrumentation::MethodProfiler.stop
+    # redis client injects a HELLO preamble on reconnection, so we will see more than 2
```
tgxworld, line 105:
```suggestion
```

This comment can be removed.

test/middleware_test.rb
```diff
@@ -90,6 +90,47 @@ def test_amzn_trace_id_in_wrong_format
...
+    redis.pipelined do |pipeline|
+      pipeline.call("PING") # => "PONG"
+      pipeline.call("PING") # => "PONG"
+    end
+
+    assert_equal(0, RedisValidationMiddleware.call_calls)
+    assert_equal(1, RedisValidationMiddleware.call_pipelined_calls)
+
+    results = PrometheusExporter::Instrumentation::MethodProfiler.stop
+    # redis client injects a HELLO preamble on reconnection, so we will see more than 2
```
tgxworld, line 130:
```suggestion
```

This comment can be removed as well.

tgxworld (approved):
@SamSaffron The changes look really good and I learnt something too :+1:

Issue #252, opened 2022-11-14
 "super:  no superclass method `async_exec'" error in v2.0.4.


snowfield702:
After upgrading to 2.0.4, the following error occurred when starting Ruby on Rails.

```
ERROR -- : super: no superclass method `async_exec' for #<PG::Connection:0x00007ff19d120b38 host=example.com port=5432 user=sample>
Did you mean?  sync_exec
               async_reset (NoMethodError)
/vendor/bundle/ruby/3.1.0/gems/prometheus_exporter-2.0.4/lib/prometheus_exporter/instrumentation/method_profiler.rb:63:in `async_exec'
```

We choose `:prepend` for the style of method patching.
https://github.com/discourse/prometheus_exporter#choosing-the-style-of-method-patching

I think that it is the effect of changing `prepend_instument` to `klass` in the following correspondence.
https://github.com/discourse/prometheus_exporter/pull/250/files#diff-cb9a335ee1d17a77e55d1bc374c4c76ebbf05e6b84de4ce41f1fbc560bda41b7R63

oakcask:
Hi maintainers,

Checked the code and wrote a patch below, I thought it would fix this
https://github.com/discourse/prometheus_exporter/pull/254 
Could you please take a time to check this up?

I and my colleagues always helped by prometheus_exporter gem. Thank you.

SamSaffron:
thanks heaps for the PR and bug report, just working through the test cases atm, will try to do a release tomorrow once sorted

PR #254, opened 2022-11-15
Fix MethodProfiler to preserve the target class behavior

oakcask:
I suspect that the change introduced by https://github.com/discourse/prometheus_exporter/pull/250/files#diff-cb9a335ee1d17a77e55d1bc374c4c76ebbf05e6b84de4ce41f1fbc560bda41b7R68 applies the patch to the original class.
It causes NoMethodError and breaks the bahavior when the original class doesn't have super class.

I think this will fix https://github.com/discourse/prometheus_exporter/issues/252 .

### DIFF
--- a/lib/prometheus_exporter/instrumentation/method_profiler.rb
+++ b/lib/prometheus_exporter/instrumentation/method_profiler.rb
@@ -64,9 +64,9 @@ class PrometheusExporter::Instrumentation::MethodProfiler
   end
 
   def self.patch_using_prepend(klass, methods, name)
-    prepend_instument = Module.new
-    define_methods_on_module(klass, methods, name)
-    klass.prepend(prepend_instument)
+    prepend_instrument = Module.new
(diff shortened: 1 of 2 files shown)


SamSaffron:
Oh nice catch! thanks for adding the test as well.

SamSaffron (approved):

SamSaffron:
getting a stack level too deep in some of the tests, can you have a look?

oakcask:
@SamSaffron

Thank you for keeping in touch.

In commit [c8976ce](https://github.com/discourse/prometheus_exporter/pull/254/commits/c8976cefa4eee2f49ed32f1c5908d83734c8294f),  put out invocation of `
  PrometheusExporter::Instrumentation::MethodProfiler.patch` from `setup` because the `MethodProfiler.patch` cannot be reverted cleanly.

SamSaffron:
cool looks good, will do a version release shortly

Issue #240, opened 2022-06-15
Sidekiq Stats Collector Memory Leak


gnomus:
Hello there,

we have Investigated an Issue with one of our prometheus_exporter Setups which stopped working.
After a certain perioud of runtime the collector started to report collector_working with Value 0 and all our Instrumentation Metrics went missing.
The only difference we found to another Setup of our which has no issues was the usage of the SideKiqStats Instrumentation so we looked into that.

It seems that [here](https://github.com/discourse/prometheus_exporter/blob/main/lib/prometheus_exporter/server/sidekiq_stats_collector.rb#L43) new observations are added to the sidekiq_metrics object which results in it getting bigger and bigger.
After a certain amount of time the Collector is not able to generate the Metric Text withing the configured default Timeout of 2 Seconds and we start seeing the above mentioned behavior.

As a comparison the [SidekiqQueueCollector](https://github.com/discourse/prometheus_exporter/blob/main/lib/prometheus_exporter/server/sidekiq_queue_collector.rb) seems to have a mechanism to clean up older Observations.

Our guess would be that the SidekiqStatsCollector would need something like that too.

NickLarsenNZ:
Does #256 close this issue?

PR #256, opened 2022-11-15
FIX: sidekiq stats collector memory leak

mpucholblasco:
We also experienced the problem explained in #240 , in which metrics present in the `sidekiq` stats collector started to grow and saturated the process (used all CPU assigned to it).

This PR aims to clean old metrics, as [sidekiq_queue_collector.rb](https://github.com/discourse/prometheus_exporter/blob/main/lib/prometheus_exporter/server/sidekiq_queue_collector.rb) does.

### DIFF
--- a/lib/prometheus_exporter/server/sidekiq_stats_collector.rb
+++ b/lib/prometheus_exporter/server/sidekiq_stats_collector.rb
@@ -40,6 +40,9 @@ module PrometheusExporter::Server
     end
 
     def collect(object)
+      now = ::Process.clock_gettime(::Process::CLOCK_MONOTONIC)
+      object['created_at'] = now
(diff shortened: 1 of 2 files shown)


SamSaffron:
Oh ... nice catch, thanks for fixing!

SamSaffron (approved):

SamSaffron:
Let me know that this resolves the issue, once confirmed I will push out a release.

mpucholblasco:
> Let me know that this resolves the issue, once confirmed I will push out a release.

Hi @SamSaffron , thanks for the fast reply. Sure, gonna test it. I will notify to you with the results.

mpucholblasco:
Hello @SamSaffron , after some tests, seems that the problem has been fixed, so I guess that you can push out a release.

Thanks!

SamSaffron:
cool, will push a new release today, thanks!

PR #257, opened 2022-11-17
Use specified labels over default labels in merge conflicts

ttstarck:
Currently, `PrometheusExporter::Metrics::Base.default_labels` will overwrite the labels specified when observing metrics in the `labels_text` output. This isn't ideal in scenarios where you want to specify a label key/value pair that doesn't match the default labels.

In this pull request, I've swapped the ordering of the merge such that we merge the specified labels into the default labels, meaning the specified labels have precedence over the default ones.


SamSaffron (approved):

SamSaffron:
Nice fix! thank you

PR #260, opened 2022-12-26
Allow binding server to both ipv4/ipv6 interfaces

sosedoff:
Starting `prometheus_exporter -v -b ALL -p 8080` will include the log message:

```
2022-12-26 16:54:44 -0600 Starting prometheus exporter on ALL:8080
[2022-12-26 16:54:44] INFO  Listening on both 0.0.0.0/:: network interfaces
[2022-12-26 16:54:44] INFO  WEBrick 1.7.0
[2022-12-26 16:54:44] INFO  ruby 2.7.2 (2020-10-01) [arm64-darwin21]
[2022-12-26 16:54:44] INFO  WEBrick::HTTPServer#start: pid=86361 port=8080
```

netstat: 

```
$ netstat -an -ptcp | grep LISTEN | grep 8080
tcp4       0      0  *.8080                 *.*                    LISTEN
tcp6       0      0  *.8080                 *.*                    LISTEN
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -857,6 +857,9 @@ prometheus_exporter -p 8080 \
                     --prefix 'foo_'
 ```
 
+You can use `-b` option to bind the `prometheus_exporter` web server to any IPv4 interface with `-b 0.0.0.0`, 
+any IPv6 interface with `-b ::`, or `-b ANY` to any IPv4/IPv6 interfaces available on your host system.
+
 #### Enabling Basic Authentication
 
 If you desire authentication on your `/metrics` route, you can enable basic authentication with the `--auth` option.
--- a/lib/prometheus_exporter/server/web_server.rb
+++ b/lib/prometheus_exporter/server/web_server.rb
@@ -44,6 +44,11 @@ module PrometheusExporter::Server
 
       @logger.info "Using Basic Authentication via #{@auth}" if @verbose && @auth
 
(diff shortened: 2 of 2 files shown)


SamSaffron:
I think @xfalcox noticed a similar issue, in the original versions we supported a bind all and then it was no longer there.

I am happy with this change, but can we update the documentation as well?

sosedoff:
Yep, i will update the docs

xfalcox:
Yes, had the same problem when updating from .5 to the latest version of this gem in Discourse.

Over there I fixed this problem by passing `*` as the bind address, which in glibc means that we will bind to both IPv4 and IPv6. However that doesn't work if someone is using musl. So 👍 for this approach from me.

sosedoff:
Should be ready to go!

SamSaffron (approved):

gphat:
*arrives late* Hey friends, any plans to release an updated version with this included? I was confused as to why it wasn't working only to release it's not in the 2.0.6 version that is most recent at this time. 😸

SamSaffron:
will try to post a new version tomorrow!

gphat:
Appreciated! I just set to pull from head, but lost some time to being confused why the `-b ANY` wasn't working for me, only to realize it wasn't released yet. 😆

Issue #261, opened 2022-12-29
Sidekiq process metrics are not being cleared out


sosedoff:
Im running my app in containers and do frequent deploys (cloud platform does not matter). Sidekiq processed are being scaled up and down based on the queue size, which means sidekiq will produce a bunch of metrics with a different `hostname`/`identity` tags. I did notice that a bunch of old metrics are still being reported in the `prometheus_exporter` metrics endpoint, though im not entirely sure if this is intentional or not.

Thought it was something to do with metrics collection in my setup, so i decided to replicate the setup locally:

1. Start `prometheus_exporter` process 
2. Start `rails server`, to get some basic web metrics in, schedule a handful of sidekiq jobs
3. Start `sidekiq`, i've used around 5 separate worker processes (each running 5 worker threads)
4. Let sidekiq workers process jobs for a few mins
5. Check metrics

```
ruby_sidekiq_process_busy{labels="",queues="default",quiet="false",tag="superapp",hostname="dans-mbp14.lan",identity="dans-mbp14.lan:3347:a2992973bcff"} 5
ruby_sidekiq_process_busy{labels="",queues="default",quiet="false",tag="superapp",hostname="dans-mbp14.lan",identity="dans-mbp14.lan:3939:6ff3ccf91542"} 5
ruby_sidekiq_process_busy{labels="",queues="default",quiet="false",tag="superapp",hostname="dans-mbp14.lan",identity="dans-mbp14.lan:7298:1a91f2ff7eeb"} 5
ruby_sidekiq_process_busy{labels="",queues="default",quiet="false",tag="superapp",hostname="dans-mbp14.lan",identity="dans-mbp14.lan:1311:14ad6d54354e"} 5
ruby_sidekiq_process_busy{labels="",queues="default",quiet="false",tag="superapp",hostname="dans-mbp14.lan",identity="dans-mbp14.lan:8404:110645868ccf"} 5

ruby_sidekiq_stats_processes_size 5
ruby_sidekiq_stats_workers_size 25
```

That looks right, all my sidekiq processes are busy.  Lets kill all but one, so we can still get sidekiq proc metrics in and simulate a scale down event. Wait a few mins to get metrics updated.

Now, checking the metrics endpoint again i still see the old worker stats, but `ruby_sidekiq_stats_processes_size` is reporting the correct process count `1`. 

[...]

PR #262, opened 2022-12-29
Handle stale sidekiq queue and process metrics

sosedoff:
This PR adds a few handling bits for dealing with stale sidekiq metics.
Root cause of the issue is described in #261.

### DIFF
--- a/lib/prometheus_exporter/server/metrics_container.rb
+++ b/lib/prometheus_exporter/server/metrics_container.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+module PrometheusExporter::Server
+  class MetricsContainer
+    METRIC_MAX_AGE = 60
+    METRIC_EXPIRE_ATTR = "_expire_at"
+
+    attr_reader :data, :ttl
+
+    def initialize(ttl: METRIC_MAX_AGE, expire_attr: METRIC_EXPIRE_ATTR)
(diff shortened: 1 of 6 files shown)


sosedoff:
With new `MetricsContainer` we can define auto-expiry upon insertion and iteration. Default is 60s, with ability to customize the value per collector.

Also, didn't want to touch other collectors, but we could expand the usage of the new class in a separate PR (if necessary).

lib/prometheus_exporter/server/metrics_container.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+
+    private
+
+    def get_time
+      ::Process.clock_gettime(::Process::CLOCK_MONOTONIC)
+    end
+
+    def wrap_expire(method_name, &blk)
+      expire.send(method_name, &blk)
```
SamSaffron, line 53: minor ... but this is a bit surprising I think, this is a bit clearer. (and more verbose) 

```
def wrap_expire(method_name, &blk)
  expire
  @data.public_send(method_name, &blk)
end
```

SamSaffron (approved):

SamSaffron:
overall change looks good :confetti_ball:  

would love an extra PR to expand usage.

SamSaffron:
Thanks @sosedoff !

PR #263, opened 2023-01-12
Add ability to filter out duplicate data in metrics container

sosedoff:
The change will allow us to use the metrics container in collectors that require checking for duplicate data.


SamSaffron:
Do we need to add anything to the readme, or is the intent just to land a change part way?

sosedoff:
This is a purely internal change for now

SamSaffron (approved):

PR #267, opened 2023-01-17
Add dockerfile

sosedoff:
This would help running the exporter service in Docker, using the official image so folks dont have to make their own.
I will also include CI/CD workflow for releasing new multi platform docker images (when a new gem version is release) if the proposal is sound.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -5,8 +5,12 @@ on:
     tags:
       - "v*"
 
+permissions:
+  contents: read
+  packages: write
+
 env:
-  DOCKER_REPO: discourse/prometheus_exporter
+  DOCKER_REPO: ghcr.io/discourse/prometheus_exporter
 
 jobs:
   build-and-publish:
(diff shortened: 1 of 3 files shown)


SamSaffron:
One concern here is that we will need to remember to update version number in 2 places ... can we just lean on version.rb somehow? Maybe we add the file... parse the version out?

sosedoff:
The goal is to have a totally automated workflow for pushing docker images - when a new tag is made, github will first trigger the current gem release workflow, and once its complete and passing, we'll trigger the docker release workflow automatically. I didnt include the changes at first since i was not sure if you wanted to have an official image in the first place.

SamSaffron:
Oh ... nice, cool lets merge it in then once that is ready. Sounds great!

sosedoff:
This should be ready. I tested the release step via my private repo and pushed out to docker hub:

Screenshot 2023-01-18 at 4 02 46 PM

For local docker build the GEM_VERSION is not set and it should install the latest version of the gem. In Github Actions we're pulling the version from the `vX.X.X` tag.

SamSaffron:
thanks @sosedoff ... let me have a quick chat with the infra team they have been trying to keep `discourse/` pretty lean on dockerhub .... we always have alternatives like just using your personal repo... will get back to you

SamSaffron:
@sosedoff discussed internally and we would prefer to publish to https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry 

Can you adjust to suit?

sosedoff:
Yea no sweat, its super easy to swap out docker providers.

SamSaffron (approved):

SamSaffron:
@sosedoff I just tried cutting a new version but the CI never ran for pushing the image... any ideas what is up... I can see the new tag. 

Maybe cause the ci pushed the tag this is not "catching"

sosedoff:
Im not entirely sure tbh, it looks like the release is done in via `rake release` which should push tags via normal channels (ie git push ..), but i'm suspecting this is not working due to push originating from the actions run. Will repro and fix.

SamSaffron:
Thanks heaps @sosedoff

PR #268, opened 2023-01-17
Expand usage of MetricsContainer in other collectors

sosedoff:
Implements the MetricsContainer where applicable

### DIFF
--- a/lib/prometheus_exporter/server/active_record_collector.rb
+++ b/lib/prometheus_exporter/server/active_record_collector.rb
@@ -13,7 +13,12 @@ module PrometheusExporter::Server
     }
 
     def initialize
-      @active_record_metrics = []
+      @active_record_metrics = MetricsContainer.new(ttl: MAX_ACTIVERECORD_METRIC_AGE)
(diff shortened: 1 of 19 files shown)


sosedoff:
@SamSaffron any thoughts? do we need extra tests?

SamSaffron:
it looks great to me... but yeah I think we should have an integration test on each of these to ensure we did not regress. 

Kind of slack we did not have the tests to start with ... process collector at least has some coverage.

Do you think I should hold off merging till we have better coverage everywhere or should we merge and revisit tests? I am open to either.. The refactor is great.

sosedoff:
Yea i'd add extra tests before merging to prevent any potential breakage.

sosedoff:
@SamSaffron please have a look around, leave feedback. Im back to wrap this PR up since it has a lot of fixes and improvements. I've got a few more tests to add but i think its already pretty solid as is.

SamSaffron:
Nice this is looking quite tidy! finding it hard to find any things to change

SamSaffron (approved):

PR #270, opened 2023-01-26
Fix docker image publishing workflow

sosedoff:
Workflow in #267 is broken due to how Github handles workflows triggered from within another workflow: 

```
When you use the repository's GITHUB_TOKEN to perform tasks, events triggered by the GITHUB_TOKEN, 
with the exception of workflow_dispatch and repository_dispatch, will not create a new workflow run. 
```

I moved the publishing job into the same CI/CD workflow, it'll kick in once the gem publishing job is complete.
It should work the same as the original `docker.yaml` flow, except the version will come from the actual ruby file.
The only downside with this approach is that target docker images will be overwritten on pushes to master even if the gem version has not changed. Maybe we could check if image exists by pulling it first, i need to test this path to be sure.

Alternatively, we could use the existing setup and generate a new Github API token for the `discoursebot` user, that way
it'll be able to schedule a run when a new tag is pushed via `rake release` (in the gem publish step). But i know nothing about the bot and how it's being used in other projects, if any, so such approach might not be feasible. 

WDYT?

(ALSO: I'd like to do a E2E test with a fake account to make sure this really works before we merge).


SamSaffron:
yeah this feels a bit fragile cause it will not necessarily push the right code to docker, relies on a certain amount of luck. 

Would far prefer is the flow is: 

- I noticed a new tag
- oh push the tag to dockerhub

let me ping @davidtaylorhq and @CvX unless they have any ideas off top of head here. I want to make sure that anything we are pushing to docker is a reproducible build

davidtaylorhq:
> Would far prefer is the flow is: I noticed a new tag

This is the logic that publish-rubygems uses for the version tagging and release:

https://github.com/discourse/publish-rubygems-action/blob/b89928239fc68d769527dd092b093b70c0e8a969/entrypoint.sh#L15-L38

Perhaps we could add a boolean output to that publish-rubygems workflow like "did_publish". Then the Docker image publishing job could be configured like:

```
needs: publish
if: ${{ needs.publish.outputs.did_publish }}
```

What do you think @cvx?

sosedoff:
We could also use the existing output `new_version` (https://github.com/discourse/publish-rubygems-action/blob/b89928239fc68d769527dd092b093b70c0e8a969/entrypoint.sh#L23) . Basically, we run the docker image publishing job if the gem publishing job was successful and `new_version` is true.

sosedoff:
@SamSaffron take a look. The mode to the gem publisher step has been implemented here https://github.com/discourse/publish-rubygems-action/pull/9, and latest PR commit addresses these modifications.

davidtaylorhq (approved):
Looks great, thanks for your work on this @sosedoff

Issue #275, opened 2023-03-23
Was the Puma memory leak fix intentionally "reverted"?


adamk9k:
Hey folks, I was searching for known issues of memory leaks and encountered this discussion from 2019: 

https://github.com/discourse/prometheus_exporter/issues/69

the outcome of which was this fix https://github.com/discourse/prometheus_exporter/commit/e2d75a5aa84bc71e8bfcf07d8a406e1145ed9958 for the Puma collector because it never released old objects from memory. 

To my surprise, this piece of code is not there in the most recent versions https://github.com/discourse/prometheus_exporter/blob/main/lib/prometheus_exporter/server/puma_collector.rb#L56-L58
and checking the blame the `collect` method of the Puma collector seems to be unchanged since it was originally added https://github.com/discourse/prometheus_exporter/blame/8db6dc16120492f1bb32a00d66ac5a0d5745c752/lib/prometheus_exporter/server/puma_collector.rb#L56-L58

The fix commit however wasn't completely reverted since I see `MAX_PUMA_METRIC_AGE` still defined as a constant (was added with the fix) but it's not used anywhere. 

Was this intentional or just a mistake? I would be happy to add it back with another PR if it's the latter. 

Thanks

adamk9k:
Ok, I will add a PR to add it back

PR #276, opened 2023-03-23
FIX: Memory leaks in unicorn + puma collectors

adamk9k:
This cherry-picks an old commit that somehow got removed/overwritten. See related discussion at: https://github.com/discourse/prometheus_exporter/issues/275


sosedoff:
MetricsCollector uses a default expiry of 60s, i think i forgot to include the usage of the puma metrics ttl constant in the PR, but i do have this addressed in the WIP PR #268

adamk9k:
Thanks, good to know. I'll close this one then.

PR #285, opened 2023-07-07
feature: add latency to delayed job metrics

martinramirez7:
This PR adds latency for delayed job jobs

### DIFF
--- a/lib/prometheus_exporter/instrumentation/delayed_job.rb
+++ b/lib/prometheus_exporter/instrumentation/delayed_job.rb
@@ -31,6 +31,7 @@ module PrometheusExporter::Instrumentation
     def call(job, max_attempts, enqueued_count, pending_count, *args, &block)
       success = false
       start = ::Process.clock_gettime(::Process::CLOCK_MONOTONIC)
+      latency = Time.current - job.run_at
       attempts = job.attempts + 1 # Increment because we're adding the current attempt
       result = block.call(job, *args)
       success = true
@@ -44,6 +45,7 @@ module PrometheusExporter::Instrumentation
         queue_name: job.queue,
         success: success,
         duration: duration,
+        latency: latency,
         attempts: attempts,
         max_attempts: max_attempts,
         enqueued: enqueued_count,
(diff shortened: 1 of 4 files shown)


martinramirez7:
Hey @SamSaffron, would you take a quick look of this please? thanks in advance!

SamSaffron:
can you also amend the readme here? Is the latency going to be accurate, can you explain the value of the metric? Also looks like CI is failing

martinramirez7:
we are trying to implement queues based on latency, the whole idea is explained [here](https://github.com/dmagliola/happy_queues). The main concept is to have queues that ensures that all of their jobs will start running in no less time than certain latency. This metric allows to detect if an increase in CPU or number of workers  is needed for a certain queue.

> can you also amend the readme here?

where exactly?

SamSaffron:
> where exactly?

REAMDE.md in the root :)

martinramirez7:
I already included the metric in the readme I'm not sure if you wanted me to add anything else. About the accuracy, there might be some loss of precision because of the sleep_delay parameter implemented by delayed job I am mentioning it in the README too.
Also, all the workflows should be passing now, would you please run the tests again?

SamSaffron:
I think we had some internal examples of time mocking without timecop, can we move to that, after that is done I am happy to merge!

martinramirez7:
Thanks Sam! Timecop removed

SamSaffron (approved):

SamSaffron:
looks good, thanks!

lib/prometheus_exporter/instrumentation/delayed_job.rb
```diff
@@ -31,6 +31,7 @@ def initialize(client: nil)
     def call(job, max_attempts, enqueued_count, pending_count, *args, &block)
       success = false
       start = ::Process.clock_gettime(::Process::CLOCK_MONOTONIC)
+      latency = Time.current - job.run_at
```
sosedoff, line 34: Is it safe to call `Time.current` here because delayed job depends on active support?

martinramirez7: Exactly, they both handle the time in the same way, DJ uses `Time.zone` within `db_time_now` method for setting the `run_at` value.

PR #287, opened 2023-07-24
ACTUALLY fix broken readme toc link

jstern:
In #286, I attempted to fix a broken readme TOC link, but I didn't check the result closely and left it still broken 🤦 The correct anchor is `histogram---custom-buckets` (3 hyphens).


CvX (approved):

CvX:
Thanks! 😃

PR #288, opened 2023-08-01
Remove post install message

DawidJanczak:
v0.5 has been released in 2020, we probably don't need this message anymore :)


SamSaffron:
good point :) thanks

SamSaffron (approved):

PR #294, opened 2023-11-14
Add missing good_job require

michaelshmitty:
It looks like a require is missing for enabling good_job instrumentation, resulting in the following error when running `bundle exec good_job start`:

`uninitialized constant PrometheusExporter::Instrumentation::GoodJob (NameError)`

This PR should fix that.


SamSaffron:
aha, we should add a test for this as well. thanks!

SamSaffron (approved):

PR #299, opened 2024-01-07
Version bump

SamSaffron:
we have been overdue a new version for a while


Appraisals
```diff
@@ -1,10 +1,13 @@
...
-  # gem "activerecord", "~> 6.0.0"
+  gem "activerecord", "~> 6.0.0"
 end
 
 appraise "ar-61" do
   gem "activerecord", "~> 6.1.1"
 end
+
+appraise "ar-70" do
+  # latest version
```
lis2, line 12:
```suggestion
  gem "activerecord", "~> 7.1.2"
```

Don't we need it? Alternatively, I think we can bump `activerecord` version in gemspec

https://github.com/discourse/prometheus_exporter/blob/5c994acd1eb9f96f3dcc71c8cdf8cf41490b11e8/prometheus_exporter.gemspec#L39

lis2 (approved):

Appraisals
SamSaffron: no idea I was just fiddling with buttons till I got it working :) I guess we need to add 71 and 70 to the matrix

PR #300, opened 2024-01-09
Add marking and sweeping time as Process stat

benoittgt:
Add marking and sweeping time as GC stats. It was introduced in https://bugs.ruby-lang.org/issues/19437

I think we should also add `GC.latest_gc_info[:major_by]`. I can do it in another PR.
```diff
--- a/lib/prometheus_exporter/instrumentation/process.rb
+++ b/lib/prometheus_exporter/instrumentation/process.rb
@@ -65,6 +65,7 @@ module PrometheusExporter::Instrumentation
       metric[:allocated_objects_total] = stat[:total_allocated_objects]
       metric[:marking_time] = stat[:marking_time]
       metric[:sweeping_time] = stat[:sweeping_time]
+      metric[:last_major_by] = GC.latest_gc_info[:major_by]
```
We could store it as a last state like kube-state-metrics does for [pod last state](https://github.com/kubernetes/kube-state-metrics/blob/90bd24f27c618ce916db4cd9da34d854d49db5b6/docs/pod-metrics.md?plain=1#L24).
It's a fast method. At the moment it's not possible because gauge are only numeric in prometheus_exporter.

cc @mtparet @byroot

### DIFF
--- a/README.md
+++ b/README.md
@@ -373,8 +373,8 @@ end
 | Counter | `major_gc_ops_total`      | Major GC operations by process               |
 | Counter | `minor_gc_ops_total`      | Minor GC operations by process               |
 | Counter | `allocated_objects_total` | Total number of allocated objects by process |
-| Gauge   | `marking_time`            | Marking time spent                           |
-| Gauge   | `sweeping_time`           | Sweeping time spent                          |
+| Gauge   | `marking_time`            | Marking time spent (Ruby 3.3 minimum)        |
+| Gauge   | `sweeping_time`           | Sweeping time spent (Ruby 3.3 minimum)       |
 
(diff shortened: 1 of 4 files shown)


SamSaffron:
should we only add the matrics on Ruby versions that have it?

benoittgt:
@SamSaffron Yes it's a good idea. I made the change. Do we want to test this?

benoittgt:
Thanks @SamSaffron. I've rebased.

SamSaffron:
thanks heaps

SamSaffron (approved):

benoittgt:
@SamSaffron Do you plan to do a new release soon ? ;)

SamSaffron:
sure, will try one today :)

PR #302, opened 2024-01-23
Add delayed_jobs_ready to DelayedJobs plugin and collect_by_queue option for GoodJob plugin

benngarcia:
This PR was born out of metric gathering for our auto-scaling needs as we're migrating hosting platforms and mid-migration of queue libraries.

This PR adds a new metric to the DelayedJobs plugin - "delayed_jobs_ready". This can be thought of as all of the jobs whose `run_at < now()`. We needed this metric and not queued or pending since those included all of our jobs which could be days, weeks, or months out.

This PR also adds the ability to view GoodJob metrics sliced by queue, similar to the DelayedJobs plugin. It's fairly self-explanatory why scaling queue workers based off how many jobs are enqueued in a given queue may be beneficial.

### DIFF
--- a/README.md
+++ b/README.md
@@ -538,6 +538,7 @@ end
 | Counter | `delayed_jobs_total`                      | Total number of delayed jobs executed                              | `job_name` |
 | Gauge   | `delayed_jobs_enqueued`                   | Number of enqueued delayed jobs                                    | -          |
 | Gauge   | `delayed_jobs_pending`                    | Number of pending delayed jobs                                     | -          |
+| Gauge   | `delayed_jobs_ready`                      | Number of ready delayed jobs                                       | -          |
 | Counter | `delayed_failed_jobs_total`               | Total number failed delayed jobs executed                          | `job_name` |
 | Counter | `delayed_jobs_max_attempts_reached_total` | Total number of delayed jobs that reached max attempts             | -          |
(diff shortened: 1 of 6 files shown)


benngarcia:
> One question, how would you handle that the number of jobs returned can decrease, because of clean up scripts? I am using the GoodJob part now, but since is a total count in the DB, the number is really unusable, when the clean up script is running next to it.

I'm not sure I fully understand the question/problem statement here - any clarification would be great :D

If you're asking about GoodJob's auto-clean up which deleted jobs after X amount of time (default 2 weeks) you can either disable the clean-up and leave the records in the db or implement some good_job on delete hook to increment some counter somewhere. Though, I'm not sure if that's within the scope of the prometheus_exporter gem, or my PR, so maybe I'm misunderstanding the question 😅

tgxworld:

benngarcia:
Hi tgxworld, no worries, life is busy :)

We would still prefer to have these changes upstreamed, and I believe most users/future users of the library would to. Each change aligns the respective plugins to have functionality that the other plugins have, creating a more consistent library experience regardless of the background job processor.

tgxworld:
@benngarcia I have reopened the PR. Can you rebase it for me to review? Thank you!

benngarcia:
@chrisrohr Good catch! We don't keep failed jobs around, and our running jobs is always capped lower than the threshold that we autoscale at - so our systems never had issues with that. But you're spot on with understanding the original intention :)

Will address this & rebase today.

Issue #301, opened 2024-01-12
Docker image not built for latest release


sosedoff:
Looks like the docker image build was skipped per https://github.com/discourse/prometheus_exporter/actions/runs/7441197605

Not sure exactly why, but do you think it'll be possible to publish the amd64/arm64 images manually?
Maybe have an option to trigger publish action instead of waiting for a new gem release.

SamSaffron:
honestly I am not sure at all, I did not build this part, not even sure how
it is all wired.

PR #303, opened 2024-01-24
Fix Github Actions workflow for publish_docker

pcallewaert:
Hi 👋 

This PR fixes 2 issues with the current workflow:
- Use v4 of the checkout action to remove the deprecation warnings about nodejs12
- Fix a typo in the output of the publish-rubygems-action. In the action it uses `new_version` as output (src: https://github.com/discourse/publish-rubygems-action/blob/v2/entrypoint.sh#L25) but we were checking on `new_version_published`. This meant the if statement for the `publish_docker` job was always false.

It's not that easy to test, but I tested this out on my fork (with some workarounds) and the `publish_docker` job triggered now successfully: https://github.com/pcallewaert/prometheus_exporter/actions/runs/7637020510

This should fix https://github.com/discourse/prometheus_exporter/issues/301

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
         activerecord: [60, 61, 70]
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
 
       - uses: ruby/setup-ruby@v1
         with:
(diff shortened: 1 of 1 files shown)


sosedoff:
Interesting. When i worked on the release flow i also used my own fork for testing, and it was working initially. Suspecting there were slight changes in the action that we missed since no new versions of this gem were published for a while (about a year?).

SamSaffron:
I am more than happy to try it out!

SamSaffron (approved):

PR #304, opened 2024-01-30
Reuse logger to log start banner instead of puts method

n-rodriguez:
Thank you!


SamSaffron:
sure seems fine, thanks

SamSaffron (approved):

n-rodriguez:
Thank you!

Issue #292, opened 2023-09-18
Handling invalid names and labels


PericlesTheo:
👋 people,

Recently we run into a scenario where a metric name we created contained [invalid characters](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels), specifically it contained a hyphen such as `"http_response_#{external_name}"` where `external_name` is an input and in this case it was `face-book` let's say.

Is there a way to handle this within the gem?

PR #310, opened 2024-03-07
Validate naming

PericlesTheo:
👋 everyone,

Related to https://github.com/discourse/prometheus_exporter/issues/292 I've opened an MR that validates the [naming expectations](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels) of Prometheus before sending the event.

I'm returning `false` for now but I am open to ideas as to the best way to handle this.

### DIFF
--- a/lib/prometheus_exporter/client.rb
+++ b/lib/prometheus_exporter/client.rb
@@ -5,7 +5,9 @@ require "logger"
 
 module PrometheusExporter
   class Client
-    class RemoteMetric
+    NAMING_VALIDATION_REGEX = /^[a-zA-Z_][a-zA-Z0-9_]*$/
+
+   class RemoteMetric
       attr_reader :name, :type, :help
 
       def initialize(name:, help:, type:, client:, opts: nil)
@@ -128,6 +130,8 @@ module PrometheusExporter
     end
 
     def send_json(obj)
+      return false unless valid_naming?(obj)
+
       payload =
         if @custom_labels
           if obj[:custom_labels]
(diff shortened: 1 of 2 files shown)


lib/prometheus_exporter/client.rb
```diff
@@ -128,6 +130,8 @@ def find_registered_metric(name, type: nil, help: nil)
     end
 
     def send_json(obj)
+      return false unless valid_naming?(obj)
```
benoittgt, line 133: Should we raise instead? Otherwise the error would be pretty silent.

Also I am wondering if could be done somewhere else to avoid extra check on every `send_json`. For example when the metric is setup?

PericlesTheo: My initial thought was to add it in `RemoteMetric` as such

```ruby
class RemoteMetric
  def valid?
     # ...
   end
end
```

But that doesn't cover the custom labels passed to the instance methods of `RemoteMetric` nor the global `custom_labels`.

Where were you thinking it would be a nice place?

About the error, I'm not sure 🤔 My initial thoughts were:
- Let's say you are using something like Sentry. You can flood your account with error events quite easily.
- Is it better to have an app crashing in the middle because of metrics?

But I'm up for discussing this.

benoittgt: You are correct. I think your change is the way to go.

SamSaffron:
this is a massive behavior change that is breaking, maybe we make this optional default off?

PericlesTheo:
Hey @SamSaffron, how/where do you suggest we do this?

Would an optional argument in `lib/prometheus_exporter/client.rb` work?

```ruby
module PrometheusExporter
  class Client
      def initialize(..., validate_metric_naming: false)
  end
end
```

If we go with an approach like that, should we raise an error when the naming is invalid?

SamSaffron:
I worry about breaking changes, I guess `validate_metric_naming: true` could be documented and use "metric_name_invalid" or something like that on invalid metrics.

PR #311, opened 2024-03-20
Update README.md - Add metrics setup instructions for puma single mode

oozzal:
There are instructions for puma single mode in other sections of the Readme. It was missing for puma core metrics.


benoittgt:
Isnt' something that should handled by the code itself?

oozzal:
@benoittgt existing readme instruction only works for Puma clustered mode and actually fails for Puma single mode.

benoittgt:
Good point

SamSaffron:
thanks!

SamSaffron (approved):

Issue #312, opened 2024-03-28
Incorrect keyword arguments handling in MethodProfiler


simpl1g:
Patched method that has keyword arguments can raise error
```ruby
require 'bundler/inline'

gemfile do
  gem 'prometheus_exporter', '2.1.0'
end

require 'prometheus_exporter/instrumentation/method_profiler'

class Test
  def method(sql, cache: false, **settings)
    puts "sql -> #{sql}"
    puts "cache -> #{cache}"
    puts "settings -> #{settings}"
  end
end

PrometheusExporter::Instrumentation::MethodProfiler.patch(Test, [:method], :some_name, instrument: :prepend)

t = Test.new

t.method('select * from table')
# sql -> select * from table
# cache -> false
# settings -> {}

t.method('select * from table', cache: true, other_setting: false)
# => wrong number of arguments (given 2, expected 1) (ArgumentError)
```

@SamSaffron I can create pull request that:
- increases minimal Ruby version to 3.0, 2.7 is already EOL for a year
- changes arguments in wrapped method to `...`, it was introduced in 2.7 => https://ryanbigg.com/2021/07/ruby-27s-new-triple-dot-syntax
`def #{method_name}(...)`

Are you ok with these changes?

PR #313, opened 2024-03-28
Fix kwargs patching

simpl1g:
Fixes https://github.com/discourse/prometheus_exporter/issues/312

### DIFF
--- a/.rubocop
+++ b/.rubocop
@@ -1,0 +1,1 @@
+--ignore-unrecognized-cops
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,3 +5,10 @@ AllCops:
   Exclude:
     - 'gemfiles/**/*'
     - 'vendor/**/*'
+
+Discourse/Plugins/NoMonkeyPatching:
+  Enabled: false
+
+Discourse/Plugins/NamespaceMethods:
+  Exclude:
+    - bin/prometheus_exporter
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ To learn more see [Instrumenting Rails with Prometheus](https://samsaffron.com/a
 
 ## Requirements
 
-Minimum Ruby of version 2.6.0 is required, Ruby 2.5.0 is EOL as of March 31st 2021.
+Minimum Ruby of version 3.0.0 is required, Ruby 2.7 is EOL as of March 31st 2023.
 
 ## Migrating from v0.x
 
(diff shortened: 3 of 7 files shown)


simpl1g:
Also it looks like in Ruby 3.2/3.3 prepend works with the same speed as alias
I did some comparison here https://gist.github.com/simpl1g/e34c073a8e33128d1f309873d9164967 and even with empty `work` there is no difference in speed. Maybe it is time to make `prepend` as default instrumentation, because it gives more benefits.

```
Comparison:
              method:  2189776.7 i/s
   method_triple_dot:  2178433.1 i/s - same-ish: difference falls within error
  prepend_triple_dot:  2161501.4 i/s - same-ish: difference falls within error
      method_prepend:  2135422.1 i/s - same-ish: difference falls within error
```

SamSaffron:
I think I am fine with this ... crazy that 2.7 was EOL a year ago!

CI is failing though, can you have a look?

simpl1g:
Thank you for review!

Looks like CI is failing for a long time, it is connected with `rubocop-discourse` gem, it is installing different version for different gemfiles. And in new version there are some changes in cops which break CI. So I made changes to rubocop config

```
BUNDLE_GEMFILE=gemfiles/ar_60.gemfile bundle list | grep rubocop-discourse
* rubocop-discourse (3.6.1)

BUNDLE_GEMFILE=gemfiles/ar_61.gemfile bundle list | grep rubocop-discourse
* rubocop-discourse (3.7.1)
```

Also ar_70.gemfile actually didn't test 7.0, but 6.0. Fixed this as well

```
BUNDLE_GEMFILE=gemfiles/ar_70.gemfile bundle list | grep active
  * activemodel (6.0.6.1)
  * activerecord (6.0.6.1)
  * activesupport (6.0.6.1)
```

.rubocop
```diff
@@ -0,0 +1 @@
+--ignore-unrecognized-cops
```
simpl1g, line 1: This is required for compatibility with older versions of `rubocop-discourse`

SamSaffron:
This is certainly a radical change (deliberately breaking releases prior to Ruby 3.0) , but I am going to accept it @tgxworld / @davidtaylorhq it is more correct and probably something we should pull into method profiler on Discourse. 

I guess at the end of the day people that need Ruby 2 support can just install earlier gems. 

Thanks @simpl1g

SamSaffron (approved):

PR #316, opened 2024-05-03
Fallback to detect controller and action as symbols

dlanderson:
In our Rails 7.1.3.2 app, action_dispatch.request.parameters is a hash with symbols. This adds a fallback to detect the action and controller names if they are symbols.


SamSaffron:
would be awesome if we could add a test here to catch this? especially if it was a general test that would work on any version and catch this in a side effect.

PR #318, opened 2024-05-29
 Update development dependencies

codez:
So I created a small PR to update some development dependencies in order to fix a couple of the test failures. I hope this does not add undesired changes.

Ruby 3.3 still has issues installing mini_racer that I did not analyze. Probably it runs into a job timeout. On my machine, it took about 3 (!) hours to complete the installation of libv8-node. Unfortunately, I have no experience with this gem.


SamSaffron:
yikes... that is scary ... will look at ruby 3.3 and mini racer ... thanks!

SamSaffron (approved):

PR #328, opened 2024-12-04
FIX: Ensure socket is closed when error is raised while opening socket

tgxworld:
/139110

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,8 @@
+2.2.0 - 2024-12-05
+
+- FIX: Ensure socket is closed when error is raised while opening socket
+- Feature: Add Dalli::Client memcache metrics for web_collector
+
 2.1.1 - 2024-06-19
 
 - FEATURE: improve good_job instrumentation
--- a/lib/prometheus_exporter/client.rb
+++ b/lib/prometheus_exporter/client.rb
@@ -59,7 +59,8 @@ module PrometheusExporter
       json_serializer: nil,
       custom_labels: nil,
       logger: Logger.new(STDERR),
-      log_level: Logger::WARN
+      log_level: Logger::WARN,
+      process_queue_once_and_stop: false
     )
       @logger = logger
       @logger.level = log_level
@@ -86,6 +87,7 @@ module PrometheusExporter
       @json_serializer = json_serializer == :oj ? PrometheusExporter::OjCompat : JSON
 
       @custom_labels = custom_labels
+      @process_queue_once_and_stop = process_queue_once_and_stop
     end
 
     def custom_labels=(custom_labels)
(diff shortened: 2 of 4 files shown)


test/client_test.rb
```diff
@@ -53,12 +53,26 @@ def test_standard_values
...
+    logger.level = :error
+
+    client = PrometheusExporter::Client.new(logger: logger, port: 321, once: true)
+    client.send("put a message in the queue")
+
+    assert_includes(
+      logs.string,
+      "Prometheus Exporter, failed to send message Connection refused - connect(2) for \"localhost\" port 321",
+    )
+  end
```
tgxworld, line 68: This test does not directly test that the socket is closed but it does cause the internal `close_socket!` method to be called.

lib/prometheus_exporter/client.rb
```diff
@@ -234,8 +241,7 @@ def ensure_socket!
 
       nil
     rescue StandardError
-      @socket = nil
```
tgxworld, line 237: Previously, we were just assigning the `@socket` to `nil` which causes a `TCPSocket` to remain open.

lib/prometheus_exporter/client.rb
```diff
@@ -59,7 +59,8 @@ def initialize(
       json_serializer: nil,
       custom_labels: nil,
       logger: Logger.new(STDERR),
-      log_level: Logger::WARN
+      log_level: Logger::WARN,
+      once: false
```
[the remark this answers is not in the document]
tgxworld: I updated to `process_queue_once_and_stop`

lib/prometheus_exporter/client.rb
tgxworld: No the worker loop is meant to ensure that a socket is opened and reuse the socket across loops. Is there a reason you think we should close the socket each time we process a message?

test/client_test.rb
tgxworld: Because it is an internal implementation detail and there is basically no way for us to assert for that in the test.

lib/prometheus_exporter/client.rb
tgxworld: Good catch and I've made the change to call `close_socket!`. However, I don't think we can easily test for this and I think it is OK since the risk of dropping a message is low.

tgxworld:
Thank you for reviewing @nattsw 👍

CHANGELOG
```diff
@@ -1,3 +1,8 @@
+2.2.0 - 2024-12-05
+
+- FIX: Ensure socket is closed when error is raised while opening socket
+- Feature: Add Dalli::Client memcache metrics for web_collector
```
nbianca, line 4: Nitpicky:

```
- FEATURE: Add Dalli::Client memcache metrics for web_collector
```

PR #337, opened 2025-04-08
Adds handling for malformed metrics

baweaver:
In the Ruby GraphQL gem we have observed some interesting behavior where collectors appear not to be registered, which causes the code from that gem to error:

```ruby
@trace.prometheus_client.send_json(
  # [!!] No name here
  type: @trace.prometheus_collector_type,
  duration: duration,
  platform_key: event_name,
  key: keyword
)
```

Source: https://github.com/rmosolgo/graphql-ruby/blob/ddf2550a204be69ba681739b529324a074d72c91/lib/graphql/tracing/prometheus_trace.rb#L68

On one hand this appears to be a malformed request body as-is, and the tests do not appear to test the integration in validating this behavior.

Given that, that brings up a potential edge case which may warrant some discussion: What should PrometheusExporter do in a case where it gets a malformed metric that does not have a name? Should it give an error? Right now what it returns is this:

```ruby
#nomethoderror: undefined method `observe' for nil:NilClass - /opt/ruby3.0/lib/ruby/gems/3.0.0/gems/prometheus_exporter-0.8.1/lib/prometheus_exporter/server/collector.rb:55:in `block in process_hash'
```

...which does not give clear traceability into the cause and potential resolutions.

The opinion of this PR, at least, is to allow it to gracefully fail when a metric cannot be registered but I am not convinced that this is the correct behavior in this scenario and would defer to the folks working on the project as to what makes the most sense here.

My suspicion with the upstream issue is that it's some form of race condition in loading collectors.

Documented some things on the GraphQL gem:

https://github.com/rmosolgo/graphql-ruby/issues/5323

Still suspecting some type of loading or race condition in here.

### DIFF
--- a/lib/prometheus_exporter/server/collector.rb
+++ b/lib/prometheus_exporter/server/collector.rb
@@ -1,8 +1,13 @@
 # frozen_string_literal: true
 
+require "logger"
+
 module PrometheusExporter::Server
   class Collector < CollectorBase
-    def initialize(json_serializer: nil)
+    attr_reader :logger
+
+    def initialize(json_serializer: nil, logger: Logger.new(STDERR))
+      @logger = logger
       @process_metrics = []
       @metrics = {}
       @mutex = Mutex.new
@@ -77,7 +82,7 @@ module PrometheusExporter::Server
       opts = symbolize_keys(obj["opts"] || {})
 
       if !name
-        STDERR.puts "failed to register metric due to empty name #{obj}"
+        logger.warn "failed to register metric due to empty name #{obj}"
         return
       end
 
@@ -96,7 +101,7 @@ module PrometheusExporter::Server
       if metric
         @metrics[name] = metric
       else
-        STDERR.puts "failed to register metric #{obj}"
+        logger.warn "failed to register metric #{obj}"
       end
     end
 
--- a/lib/prometheus_exporter/server/collector.rb
+++ b/lib/prometheus_exporter/server/collector.rb
@@ -40,6 +40,8 @@ module PrometheusExporter::Server
           metric = @metrics[obj["name"]]
           metric = register_metric_unsafe(obj) if !metric
 
+          next unless metric
+
           keys = obj["keys"] || {}
           keys = obj["custom_labels"].merge(keys) if obj["custom_labels"]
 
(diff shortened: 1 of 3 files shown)


baweaver:
One other potential strategy that would additionally remove the `ObjectSpace` invocations would be to use `self.inherited` for a registry instead, but that'd have some consequences:

* `server.rb` would be more order-dependent
* Given the usage of `mutex` it may not be great to `CollectorChild.new` and cache that in an effective singleton

...but it'd also make it easier for any downstream consumers making custom collectors.

If that is of interest I can get a PR up on that later.

lib/prometheus_exporter/server/collector.rb
```diff
@@ -74,6 +76,11 @@ def register_metric_unsafe(obj)
       help = obj["help"]
       opts = symbolize_keys(obj["opts"] || {})
 
+      if !name
+        STDERR.puts "failed to register metric due to empty name #{obj}"
```
SamSaffron, line 80: this feels risky, we should not be printing to stderr, we have a logger can we use it?

lib/prometheus_exporter/server/collector.rb
```diff
@@ -1,8 +1,13 @@
 # frozen_string_literal: true
 
+require "logger"
```
baweaver, line 3: Not sure this is needed, can remove.

lib/prometheus_exporter/server/collector.rb
```diff
@@ -1,8 +1,13 @@
...
+
 module PrometheusExporter::Server
   class Collector < CollectorBase
-    def initialize(json_serializer: nil)
+    attr_reader :logger
+
+    def initialize(json_serializer: nil, logger: Logger.new(STDERR))
```
baweaver, line 9: Added a defaulted logger targeted to STDERR to match what behavior was here before, can also change this. By default it's passing in the server logger though in the main flows.

test/server/collector_test.rb
```diff
@@ -100,6 +100,45 @@ def test_register_metric
...
+
+    collector.process(json)
+
+    assert_equal("", collector.prometheus_metrics_text)
+  end
+
+  def test_default_logger
+    collector = PrometheusExporter::Server::Collector.new
+    assert_instance_of(Logger, collector.logger)
+  end
```
baweaver, line 122: Probably a hair overkill of a test, can delete.

PR #343, opened 2025-05-29
Take delayed_job job_name from job

vpolozov:
## Problem
 The current DelayedJob instrumentation extracts job names using only a regex pattern, 
 which often results in empty job names even when the actual job name is available.

 ## Solution
 Improved the job name extraction by:
 - First attempting the original regex approach
 - Falling back to get the name from payload_object  if needed

## Details
`::Delayed::Job` has `name` method which calls `payload_object.display_name`  See `<delayed_job>/lib/delayed/backend/base.rb`


SamSaffron (approved):

SamSaffron:
thanks, yeah worth a shot

PR #350, opened 2025-08-09
Secure PrometheusExporter scraping with SSL

n-rodriguez:
Thank you!

### DIFF
--- a/lib/prometheus_exporter/client.rb
+++ b/lib/prometheus_exporter/client.rb
@@ -61,7 +61,10 @@ module PrometheusExporter
       custom_labels: nil,
       logger: Logger.new(STDERR),
       log_level: Logger::WARN,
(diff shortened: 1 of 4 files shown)


lib/prometheus_exporter/client.rb
```diff
@@ -2,6 +2,7 @@
 
 require "socket"
 require "logger"
+require "openssl"
```
SamSaffron, line 5: in web server we delay require it, I guess we should do the same here?

n-rodriguez: @SamSaffron done!

n-rodriguez:
> oh oops looks like we have a conflict cause I merged in the wrong order, can you fix

@SamSaffron done! thank you!

PR #372, opened 2026-05-22
Support EC keys in build_ssl_context

n-rodriguez:
Thank you!


n-rodriguez:
ping @SamSaffron thank you!

SamSaffron:
I would like to hold off here, we are looking to move us off webrick either to async-http or puma ...

n-rodriguez:
> I would like to hold off here, we are looking to move us off webrick either to async-http or puma ...

I understand but this is a very small change that keeps compatibilty

n-rodriguez:
ok, I see https://github.com/discourse/prometheus_exporter/pull/376

PR #375, opened 2026-07-21
DEV: replace WEBrick with a bounded socket server

### LINKED ISSUE #374: CVE-2026-38969
According to the official documentation of webrick, the gem is not supposed to be used in production and the prometheus_exporter gem directly adds the webrick as a dependency. This is flagging bundler-audit checks in the CI for many people. Refer to the official issue [here](https://github.com/ruby/webrick/pull/199).

sam-saffron-jarvis:
## Summary

Replace WEBrick with a purpose-built HTTP/1 server using Ruby's standard-library sockets and remove the `webrick` runtime dependency.

This preserves the existing public `WebServer` API and supported behavior:

- `/metrics`, including gzip and crypt htpasswd Basic Auth
- streaming chunked `/send-metrics` uploads from `PrometheusExporter::Client`
- `/ping`
- TLS
- IPv4, IPv6, `ANY`, and `ALL` binding
- concurrent clients and bounded shutdown

The currently reported `CVE-2026-38969` was rejected, but WEBrick's maintainers explicitly discourage production use. This change removes that production dependency rather than swapping it for a native-extension server dependency.

Closes #374.

## Bounds and protocol

The server is deliberately narrow and defaults to:

- **100** simultaneous connections
- **64 KiB** total request headers
- **1 MiB** per metrics chunk
- **5 seconds** for TLS negotiation and request headers
- a fresh **30-second** read deadline for each complete body chunk, longer than the client's **25-second** maximum socket age
- **5 seconds** for nonblocking response writes

It serves one request per connection with fixed `Content-Length` and `Connection: close`. It rejects ambiguous framing, duplicate/folded headers, HTTP/1.0 transfer encoding, request pipelining, chunk extensions/trailers, and unsupported transfer codings. Worker saturation leaves connections in the kernel backlog instead of accepting and dropping producer uploads.

## Verification

- Focused server suite: **35 runs, 484 assertions, 0 failures**
- Full suite: **189 runs, 812 assertions, 0 failures**
- Focused suite repeated across 10 seeds
- Full suite repeated across 3 seeds
- Syntax Tree: clean
- RuboCop: clean
- `gem build`: successful, with no runtime dependencies
- `git diff --check`: clean

Coverage includes raw malformed/ambiguous HTTP, delayed chunk reuse, buffered chunks with slow collectors, worker-cap admission and reclamation, transient accept failures, read/write deadlines, clean shutdown, gzip negotiation, auth, TLS certificate validation, bundled TLS-client uploads larger than 16 KiB, IPv6/dual-stack binding, and JSON/Oj/default client integration.

## Fable review

Fable reviewed the complete diff and found one potentially blocking issue: `require "base64"` would introduce an undeclared bundled-gem dependency on Ruby 3.4+. Fixed in `ad331bc` by using strict `String#unpack1("m0")` decoding instead. Fable's remaining observations were documented intentional protocol tightening or non-blocking lifecycle notes.

## Internal design

The initial implementation was deliberately refactored before review: the former 991-line `web_server.rb` is now three cohesive internal units:

- `web_server.rb` — exporter routes, metrics, authentication, compression, and public API (**371 lines**)
- `http_connection.rb` — bounded HTTP parsing, body framing, and response I/O (**376 lines**)
- `socket_listener.rb` — binding, TLS, admission, workers, and shutdown (**397 lines**)

The 35 server tests were split across matching route, protocol, and listener suites while retaining all **484 assertions**. A post-refactor developer review found no behavior drift or actionable issues.


SamSaffron:
puma at: https://github.com/discourse/prometheus_exporter/pull/376

PR #129, opened 2020-09-10
Move sidekiq_job_duration_seconds from Counter to Summary

dmcaulay:
I'd love to be able to see p99, p90, p50 for Sidekiq job durations.

PR #379, opened 2026-08-10
Build shoryuken_job_duration_seconds from default_aggregation

Eulisha:
# Build shoryuken_job_duration_seconds from default_aggregation

`shoryuken_job_duration_seconds` is the only job-duration metric still hardcoded to `Counter`, so it can never report anything but a mean, and the `--histogram` flag has no effect on it.

Every sibling collector builds this metric from `Base.default_aggregation`:

| Collector | duration metric |
| --- | --- |
| `web_collector.rb` | `Base.default_aggregation` |
| `sidekiq_collector.rb` | `Base.default_aggregation` |
| `delayed_job_collector.rb` | `Base.default_aggregation` |
| `shoryuken_collector.rb` | **`Metric::Counter`** |

This looks like an oversight rather than a decision. The Shoryuken collector was added in #104 (Jan 2020) by copying the Sidekiq collector as it stood then, which also used `Counter`. It was not carried along by either of the two later changes:

- #129 (Sep 2020) moved `sidekiq_job_duration_seconds` off `Counter` — "I'd love to be able to see p99, p90, p50 for Sidekiq job durations."
- #186 (Sep 2021) introduced `--histogram` and switched web + sidekiq to `Base.default_aggregation`, resolving #92.

`shoryuken_collector.rb` has had no functional change since Jul 2020.

This mirrors #186's approach rather than #129's, so the metric follows whatever the server was started with: `Summary` by default, `Histogram` under `-g`. Behaviour for users who do not pass `-g` changes from a bare counter to a summary, which is the same transition Sidekiq users saw in #129.

The three remaining metrics stay counters, as they are genuine monotonic event counts.

## Test plan

Updated `test_it_can_collect_shoryuken_metrics_with_custom_lables` to assert the quantile series, matching the assertions #129 added for Sidekiq.

Verified locally against 2.3.1 with the equivalent patch applied:

```
# default (Summary)
shoryuken_job_duration_seconds{job_name="String",queue_name="",service="service1",quantile="0.99"} 0.05
...
shoryuken_job_duration_seconds_sum{job_name="String",queue_name="",service="service1"} 0.05
shoryuken_job_duration_seconds_count{job_name="String",queue_name="",service="service1"} 1

# with -g (Histogram), observations 0.02 / 0.05 / 0.05 / 0.06 / 3.5
# TYPE shoryuken_job_duration_seconds histogram
shoryuken_job_duration_seconds_bucket{...,le="0.025"} 1
shoryuken_job_duration_seconds_bucket{...,le="0.05"} 3
shoryuken_job_duration_seconds_bucket{...,le="0.1"} 4
shoryuken_job_duration_seconds_bucket{...,le="2.5"} 4
shoryuken_job_duration_seconds_bucket{...,le="5.0"} 5
shoryuken_job_duration_seconds_bucket{...,le="+Inf"} 5
shoryuken_job_duration_seconds_count{...} 5
shoryuken_job_duration_seconds_sum{...} 3.68
```

`_sum` is unchanged, so anyone currently deriving a mean from this metric can keep doing so via `_sum` / `_count`.

## Note for users

This renames the exposed series: the bare `shoryuken_job_duration_seconds` becomes `_sum` / `_count` (plus `_bucket` under `-g`). Dashboards reading the bare name need updating. Worth a CHANGELOG entry.


Eulisha:
Hello @SamSaffron, would you have time to take a look at this one?

It's a one-line change following the same approach as #186, which you reviewed — `shoryuken` is the last collector still building its duration metric from a `Counter`, so `--histogram` currently has no effect on it.

Happy to add a CHANGELOG entry for the series rename, or to adjust the approach if you'd prefer something else here.

SamSaffron:
no worries. happy to try it.

SamSaffron (approved):

PR #380, opened 2026-08-18
fix: avoid sending empty `job_name` label for delayed job metrics

jzyeezy:
Only ActiveJob-wrapped jobs have handlers whose YAML contains a `job_class:` key.

Metrics related to delayed jobs enqueued directly through plain `delayed_job`, like `some_object.delay.some_method`, `handle_asynchronously`, or `Delayed::Job.enqueue(CustomJob.new)` will have no `job_class:` key in the job's handler YAML.

The existing code will return `nil`, which turns into an empty string `""` on `.to_s`.

This change allows the former use-case of delayed jobs to have sensible `job_name` labels on their prometheus metrics.


jzyeezy:
Hi @SamSaffron - would you mind reviewing? Thanks!

Edit - sorry for the PR noise - classic PEBKAC 😅

